PR #18, opened 2021-11-18
Onboard Component Controller for App Service

maysunfaisal:
This PR introduces the component controller for App Service

- Accepts only Github src at the moment, not container image
- Creates a Component CR and updates the corresponding Application CR
- Tests


config/samples/appstudio_v1alpha1_component.yaml
```diff
@@ -1,7 +1,29 @@
 apiVersion: appstudio.redhat.com/v1alpha1
```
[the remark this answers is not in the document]
johnmcollier: Maybe something like:
```
kind: Component
metadata:
  name: component-sample
  name: hascomponent-sample-12345
spec:
  componentName: backend
  application: hasapplication-sample-12345
  source:
    git:
      url: https://github.com/devfile-samples/devfile-sample-java-springboot-basic
```

controllers/component_controller.go
```diff
@@ -47,10 +60,197 @@ type ComponentReconciler struct {
...
+				}
+
+				endpoint := rawURL + "/" + devfilePath
+				devfileBytes, err = util.CurlEndpoint(endpoint)
+				if err != nil {
+					log.Error(err, fmt.Sprintf("Unable to curl %s, attempting to clone the repo and read the devfile instead %v", devfilePath, req.NamespacedName))
+
+					// clone the repo and read the devfile
+					clonePath := path.Join(clonePathPrefix, hasComponent.Spec.Application, hasComponent.Spec.ComponentName)
+					devfileBytes, err = util.CloneAndReadDevfile(clonePath, devfilePath, source.GitSource.URL)
```
johnmcollier, line 112: Do we clean up and remove the temp folder after we clone and retrieve the devfile? If not, we should.

controllers/component_controller.go
```diff
@@ -47,10 +60,197 @@ type ComponentReconciler struct {
...
+				hasComponent.SetLabels(hasComponentLabels)
+				err = r.Client.Update(ctx, &hasComponent)
+				if err != nil {
+					log.Error(err, fmt.Sprintf("Unable to update Component with the required labels %v", req.NamespacedName))
+					r.SetCreateConditionAndUpdateCR(ctx, &hasComponent, err)
+					return ctrl.Result{}, err
+				}
+			} else {
+				log.Error(err, fmt.Sprintf("Application devfile model is empty. Before creating a Component, an instance of Application should be created, exiting reconcile loop %v", req.NamespacedName))
+				err := fmt.Errorf("application devfile model is empty. Before creating a Component, an instance of Application should be created")
```
johnmcollier, line 209: Thinking out loud here: Should we surface some of these error messages in the `kubectl describe ...` output and also add a `Status` field to the `kubectl get hascomp` / `kubectl get hasapp` output?

What do you think? We don't need to do this in your PR, but should probably open an issue if we think it's worthwhile.

controllers/component_controller.go
```diff
@@ -47,10 +60,197 @@ type ComponentReconciler struct {
...
+				r.SetCreateConditionAndUpdateCR(ctx, &hasComponent, err)
+				return ctrl.Result{}, err
+			}
+
+		} else if source.ImageSource != nil && source.ImageSource.ContainerImage != "" {
+			log.Info(fmt.Sprintf("container image is not supported at the moment, please use github links for adding a component to an application for %v", req.NamespacedName))
+		}
+	} else {
+		// If the model already exists, see if fields have been updated
+		log.Info(fmt.Sprintf("Checking if the HAS Component has been updated %v", req.NamespacedName))
```
johnmcollier, line 219:
```suggestion
		log.Info(fmt.Sprintf("Checking if the Component has been updated %v", req.NamespacedName))
```

🙂  

There's a few other instances too

controllers/component_controller_test.go
```diff
@@ -0,0 +1,562 @@
...
+		HASAppName      = "test-hasapplication-123"
+		HASCompName     = "test-hascomponent-123"
+		HASAppNamespace = "default"
+		DisplayName     = "petclinic"
+		Description     = "Simple petclinic app"
+		ComponentName   = "backend"
+		SampleRepoLink  = "https://github.com/devfile-samples/devfile-sample-java-springboot-basic"
+	)
+
+	Context("Create Component with basic field set", func() {
```
johnmcollier, line 51: Could you add a simple case here where the devfile is malformed? Maybe by pointing `devfileURL` to a devfile that cannot be parsed into a valid devfile?

johnmcollier: And maybe another case where the `Application` CR referenced doesn't exist?

controllers/component_controller_test.go
```diff
@@ -0,0 +1,562 @@
...
+
+	requests := checklist.resources.Requests
+	limits := checklist.resources.Limits
+
+	for i, component := range components {
+		attributes := component.Attributes
+		var err error
+
+		// Check the route
+		route := attributes.Get("appstudio.has/route", &err)
```
johnmcollier, line 407: Not sure if I'm following along fully: What happens `appstudio.has/route` (and other attributes) aren't set in the devfile for a valid test case?

controllers/update.go
```diff
@@ -0,0 +1,309 @@
...
+
+	devfileAPIV1 "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
+	"github.com/devfile/api/v2/pkg/attributes"
+	data "github.com/devfile/library/pkg/devfile/parser/data"
+	"github.com/devfile/library/pkg/devfile/parser/data/v2/common"
+	appstudiov1alpha1 "github.com/redhat-appstudio/application-service/api/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+)
+
+func (r *ComponentReconciler) updateComponentDevfileModel(hasCompDevfileData data.DevfileData, hasComponent appstudiov1alpha1.Component) (bool, error) {
```
johnmcollier, line 30: This function generally looks fine to me, but I'm wondering about its long-term maintainability? Especially as we add more fields to the Component CR.

What if we generate the updated component devfile model, and compare it against the existing one (maybe via `reflect.DeepEqual` or something similar)? And if they're not equal, we update the devfile in the status to the new one?

Once we have webhooks in place, we should be able to block any invalid changes (so we shouldn't have to worry about those sneaking in).

johnmcollier (changes requested):

controllers/component_controller.go
```diff
@@ -47,10 +60,197 @@ type ComponentReconciler struct {
...
+				if err != nil {
+					log.Error(err, fmt.Sprintf("Unable to convert Github URL to raw format, exiting reconcile loop %v", req.NamespacedName))
+					r.SetCreateConditionAndUpdateCR(ctx, &hasComponent, err)
+					return ctrl.Result{}, err
+				}
+
+				endpoint := rawURL + "/" + devfilePath
+				devfileBytes, err = util.CurlEndpoint(endpoint)
+				if err != nil {
+					log.Error(err, fmt.Sprintf("Unable to curl %s, attempting to clone the repo and read the devfile instead %v", devfilePath, req.NamespacedName))
```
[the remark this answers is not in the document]
maysunfaisal: we can remove clone if we dont want it, I'll keep it simple for now and remvoe it

controllers/component_controller.go
maysunfaisal: so would it be empty most of the time if there is no err?

controllers/component_controller.go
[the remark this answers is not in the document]
johnmcollier: Let's discuss on today's call.

controllers/component_controller.go
johnmcollier: My thinking was it would be set to `Created` / `Updated` / `Error`, etc? But I dunno if that'd be useful.

controllers/component_controller_test.go
maysunfaisal: yeah currently I am setting it everytime I call this func, let me see what i can do

controllers/update.go
maysunfaisal: The thing is `reflect.DeepEqual` will not work on struct with pointers; since two structs can have the same data but those structs may have pointers referencing diff things; so it will fail

I have tried to use `reflect.DeepEqual` before going this way

maysunfaisal: let me try out a few reflect funcs and see it actually works

controllers/component_controller.go
maysunfaisal: created an [issue](https://issues.redhat.com/browse/DEVHAS-49)

controllers/update.go
maysunfaisal: This func now generates from Component CR; and we just do a `reflect.DeepEqual` to see the diff

controllers/component_controller_test.go
maysunfaisal: This test func now checks if the checklist arguments are present before checking

controllers/component_controller.go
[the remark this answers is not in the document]
maysunfaisal: Removed as discussed in the meeting

controllers/component_controller.go
johnmcollier: Sounds good, thanks. Let's discuss on today's call.

johnmcollier (approved):

PR #19, opened 2021-11-23
Use the Red Hat UBI image from Red Hat registry

sbose78:
Use registry.access.redhat.com/ubi8/ubi-minimal:8.4-210


johnmcollier (approved):
LGTM

Thanks Shoubhik, we're gonna need to be based on UBI, so this is good to get in 🙂

PR #21, opened 2021-12-06
WIP ensure container doesn't complain about root

sbose78:
.. on openshift


sbose78:
This may not be needed..

johnmcollier (approved):
@sbose78 I was going to comment, I don't think I've seen this issue on OpenShift when I've tested recently.

I've got no concerns explicitly adding `USER 65532:65532` to the Dockerfile either way, if we want to be clear. Up to you, feel free to either merge or close this PR.

johnmcollier:
I'm gonna merge this - was doing some testing on Minikube and needed to set the container to non-root

PR #22, opened 2021-12-07
"Invoke builds" upon component creation

sbose78:
Upon creation of a component with an destination image specified:

* Trigger an image build.
* Setup webhooks and print the webhook url in the `.status.webhook` of the component for the UI to render the same.

---

```
kind: Application
metadata:
  name: my-big-application
spec:
  description: application definition for petclinic-app
  displayName: petclinic
---
kind: Component
metadata:
  name: component-a
spec:
  componentName: component-a
  application: my-big-application
  source:
    git:
      url: https://github.com/sbose78/devfile-sample-code-with-quarkus
  build:
    containerImage: docker.io/sbose78/foobar:dev-code-with-quarkus
```


controllers/component_controller.go
```diff
@@ -253,8 +253,22 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		}
 	}
 
+	if component.Spec.Source.ImageSource != nil {
```
maysunfaisal, line 256: So I am not sure if our understanding of the `Component` spec is on the same page here. I took a glance at the implementation and it seems like when `component.Spec.Source.ImageSource` is specified, we build from `component.Spec.Source.GitSource.URL` and output it to the image `component.Spec.Source.ImageSource` 

But `component.Spec.Source.ImageSource` & `component.Spec.Source.GitSource.URL` are two kinds of src. A user on HAC during import; can choose to import from devfile sample/their own git project or specify an image. This image is represented as `component.Spec.Source.ImageSource`, so its an Union.

If I'm right, we are interested in a devfile's [component.image.imageName](https://github.com/nodeshift-starters/devfile-sample/blob/main/devfile.yaml#L17) which holds the output where image build needs to go to

sbose78: Ah I see, I wasn't aware of that.

sbose78: Either way, the 'output' image needs to be presented by the user - is what I think.

sbose78:
TODO: I still need the e2e tests to be a little more elaborate.

api/v1alpha1/component_types.go
```diff
@@ -106,6 +106,9 @@ type ComponentStatus struct {
 	// Condition about the Component CR
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
 
+	// Webhook URL generated by Builds
+	Webhook string `json:"webhook,omitempty"`
```
sbose78, line 110: Needs special review since this is an API change.

controllers/build.go
```diff
@@ -0,0 +1,350 @@
...
+				return "", err
+			}
+
+		} else {
+			log.Error(err, fmt.Sprintf("Unable to get build %v", eventListener))
+			return "", err
+		}
+	}
+
+	var port int32 = 8080
```
maysunfaisal, line 302: would suggest to use `component.spec.TargetPort`

maysunfaisal:
@sbose78 I've a few questions based on this PR 🤔
- who will be monitoring the webhook url in `.status.webhook` after creating it? How does the Build Service come into the picture  
- we probably need to gate the logic a bit more; if the Git src is a devfile sample, the sample repo will not be updated. So the build will be based on the cloned repo off that sample repo..
- it seems the call to `setupWebhookTriggeredImageBuilds` is done in the Component reconcile, so what does it mean when the reconcile function is invoked on a Component CR update

controllers/component_controller.go
maysunfaisal: Interesting, so will we be using an output image for every build? I was of the opinion that it would be only be for the outerloop 🤔

sbose78:
> who will be monitoring the webhook url in .status.webhook after creating it? How does the Build Service come into the picture

That logic has to be built into HAS itself to extract the correct webhook url ( assuming it has a chance to from the initial url ) 

> we probably need to gate the logic a bit more; if the Git src is a devfile sample, the sample repo will not be updated. So the build will be based on the cloned repo off that sample repo..

* User tries out with sample and is satisfied. Doesn't get to use webhooks though.
* User creates a new component with his clone and uses a webhook there. 

This is how I'm envisioning this. What do you think?

> it seems the call to setupWebhookTriggeredImageBuilds is done in the Component reconcile, so what does it mean when the reconcile function is invoked on a Component CR update

At the moment, the idea is to not constantly overwrite changes, which is why, the `setupWebhookTriggeredImageBuilds` only creates the resources if they were removed. This leaves room for us to make HAS the starting point for things but not necessarily manage it forever. The ultimate goal here would be
* push these resources to the user's git repo.
* have gitops service configure them and apply them onto the cluster.
* subsequently, the user could manage them off Git directly or using the App Studio experience - though the latter would have limited functionality only.

amfred:
A couple of maybe-silly questions from the spec:
```
spec:
  componentName: backend
  source:
    git:
      url: https://github.com/sbose78/devfile-sample-code-with-quarkus
    image:
      containerImage: docker.io/sbose78/foobar:dev-code-with-quarkus
```
Is "componentName" provided by the end user?
What is the foobar here and how does it get set?

controllers/component_controller.go
amfred: A couple of maybe-silly questions from the spec:
```
spec:
  componentName: backend
  source:
    git:
      url: https://github.com/sbose78/devfile-sample-code-with-quarkus
    image:
      containerImage: docker.io/sbose78/foobar:dev-code-with-quarkus
```
Is "componentName" provided by the end user?

What is the foobar here and how does it get set?

I see @maysunfaisal referring to component.Spec.Source.ImageSource, component.Spec.Source.GitSource.URL,  component.Spec.Source.ImageSource, and component.image.imageName.  How does those map to the field names in your PR comment: component.SpecSource.Git.URL and component.Spec.Source.Image.ContainerImage?

controllers/component_controller.go
```diff
@@ -253,8 +253,22 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		// TODO: would move this to the user's gitops repository under /.tekton
+		webhookURL, err := r.setupWebhookTriggeredImageBuilds(ctx, log, component)
+		if err != nil {
+			log.Error(err, "Unable to setup builds")
+		}
+		component.Status.Webhook = webhookURL
+		err = r.Client.Status().Update(ctx, &component)
+		if err != nil {
+			log.Error(err, "Unable to update Component with webhook URL")
+		}
```
maysunfaisal, line 267: so i think we should do this only when the webhook returned is different.

Going by your previous comment [here](https://github.com/redhat-appstudio/application-service/pull/22#issuecomment-992723798), if there is already a webhook url created, we dont necessarily want to keep updating this webhook field everytime it goes thru reconcile 🤔

maysunfaisal:
> > we probably need to gate the logic a bit more; if the Git src is a devfile sample, the sample repo will not be updated. So the build will be based on the cloned repo off that sample repo..
> 
> * User tries out with sample and is satisfied. Doesn't get to use webhooks though.
> * User creates a new component with his clone and uses a webhook there.
> 
> This is how I'm envisioning this. What do you think?
> 

yes, that seems fine to me now, i just went through the [google doc](https://docs.google.com/document/d/1IPlihVjkJ4Kb9tdhsk7iz3bn5rkT_SvJCNQhyXzK3aI/edit?pli=1#heading=h.nqdz8ulymna6) and saw the flow which makes sense.

> > it seems the call to setupWebhookTriggeredImageBuilds is done in the Component reconcile, so what does it mean when the reconcile function is invoked on a Component CR update
> 
> At the moment, the idea is to not constantly overwrite changes, which is why, the `setupWebhookTriggeredImageBuilds` only creates the resources if they were removed. This leaves room for us to make HAS the starting point for things but not necessarily manage it forever. The ultimate goal here would be
> 
> * push these resources to the user's git repo.
> * have gitops service configure them and apply them onto the cluster.
> * subsequently, the user could manage them off Git directly or using the App Studio experience - though the latter would have limited functionality only.

hmm I see

controllers/component_controller.go
maysunfaisal: @amfred I can answer few of these ques here

> I have a few maybe-silly questions from the spec in your PR comment, @sbose78 :
> 
> ```
> spec:
>   componentName: backend
>   application: application-sample
>   source:
>     git:
>       url: https://github.com/sbose78/devfile-sample-code-with-quarkus
>     image:
>       containerImage: docker.io/sbose78/foobar:dev-code-with-quarkus
> ```
> 
> Is "componentName" provided by the end user?
> 

I think yes, when a user adds the component via HAC, they would probably have a way to specify the component name that is being added.

> What is the foobar here and how does it get set?
> 

I think this is just the image name that is being called here in this example, it probably should be backend to reflect the component in question(probably namespaced to handle it better)

> I see @maysunfaisal referring to component.Spec.Source.ImageSource, component.Spec.Source.GitSource.URL, component.Spec.Source.ImageSource, and component.image.imageName. How does those map to the field names in your PR comment: component.SpecSource.Git.URL and component.Spec.Source.Image.ContainerImage?

HAC imports cases
1. devfiile sample/git - HAC uses component.Spec.Source.GitSource.URL
2. container image - HAC uses component.Spec.Source.ImageSource

In use case 1, the devfile that is associated will have a component component[i].image.imageName which can be either preset/input via HAC by the user. And this will store the build output.

In use case 2, since its only a container image(String) thats being inputted, a basic devfile would be created with the image specified. Not entirely sure how the builds would work in this case.

controllers/component_controller.go
sbose78: That is correct.

sbose78:
will consume https://github.com/redhat-appstudio/application-service/pull/26

sbose78:
From PM: 

* For Milestone 2, we said the person doing the demo can just show the container image by going into quay.io or run a curl/wget command
* The UI does not need to show it. That's in milestone 3
* The link is to the place that app studio determines for hosting the image. Not editable by the user

sbose78:
Updates:

* If `.spec.build` is left out, the controller uses `quay.io/redhat-appstudio/user-workload:${namespace}-${component-name}`.
* For event-based builds, the image would be pushed to `quay.io/redhat-appstudio/user-workload:${namespace}-${component-name}-${git-revision}`
* The registry secret needs to be placed in a `secret` called `redhat-appstudio-registry`.

This is what the controller generates:

```
apiVersion: tekton.dev/v1beta1
kind: PipelineRun
metadata:
  generateName: component-sample-
  annotations:
    build.appstudio.openshift.io/repo: 'https://github.com/sbose78/devfile-sample-code-with-quarkus'
    build.appstudio.openshift.io/type: build
    build.appstudio.openshift.io/version: '0.1'
  namespace: sbose
  labels:
    pipelines.openshift.io/runtime: generic
    pipelines.openshift.io/strategy: docker
    pipelines.openshift.io/used-by: build-cloud
spec:
  params:
    - name: git-url
      value: 'https://github.com/sbose78/devfile-sample-code-with-quarkus'
    - name: output-image
      value: >-
        quay.io/redhat-appstudio/user-workload:sbose-component-sample-local
  pipelineRef:
    bundle: 'quay.io/redhat-appstudio/build-templates-bundle:v0.1.2'
    name: devfile-build
  serviceAccountName: pipeline
  timeout: 1h0m0s
  workspaces:
    - name: workspace
      persistentVolumeClaim:
        claimName: appstudio
      subPath: component-sample/a120e8138a33d34bd0689d75883f5b02d5a51aae
    - name: registry-auth
      secret:
        secretName: redhat-appstudio-registry
```

sbose78:
```
{"reconciler group": "appstudio.redhat.com", "reconciler kind": "Component", "name": "test-application-1234", "namespace": "default", "error": "no matches for kind \"TriggerTemplate\" in version \"triggers.tekton.dev/v1alpha1\""}
```

Even though I've added the schemas, the tests don't seem to like any of the added APIs. The error message did improve upon adding them ( which means they are taking effect to a tiny extent ). 

https://github.com/redhat-appstudio/application-service/pull/22/files#diff-522ed98795d88bff1df848662d1eee2a296ed9a37b852695f82583865d198092R83

controllers/build.go
```diff
@@ -0,0 +1,263 @@
...
+
+	if strings.Contains(outputImage, ":") {
+		outputImage = outputImage + "-" + "$(tt.params.git-revision)"
+	} else {
+		outputImage = outputImage + ":" + "$(tt.params.git-revision)"
+	}
+	return outputImage
+}
+
+func paramsForInitialBuild(component appstudiov1alpha1.Component) []tektonapi.Param {
```
maysunfaisal, line 95: would you mind pls putting a small desc what the func does, so that i dont have to scroll and check for a quick glance every time i come across it

probably a good idea for all the funcs in the file

controllers/build.go
```diff
@@ -0,0 +1,263 @@
...
+			Kind:       "PipelineRun",
+			APIVersion: tektonapi.SchemeGroupVersion.Group + "/" + tektonapi.SchemeGroupVersion.Version,
+		},
+		Spec: webhookBasedBuildTemplate,
+	}
+
+	resourceTemplatePipelineRunBytes, err := json.Marshal(resoureTemplatePipelineRun)
+	if err != nil {
+		return nil, err
+		//og.Error(err, fmt.Sprintf("Unable to convert to bytes %v", resoureTemplatePipelineRun))
```
maysunfaisal, line 217: comment found

controllers/build.go
```diff
@@ -0,0 +1,263 @@
...
+		outputImage = outputImage + "-" + "$(tt.params.git-revision)"
+	} else {
+		outputImage = outputImage + ":" + "$(tt.params.git-revision)"
+	}
+	return outputImage
+}
+
+func paramsForInitialBuild(component appstudiov1alpha1.Component) []tektonapi.Param {
+	sourceCode := component.Spec.Source.GitSource.URL
+	outputImage := component.Spec.Build.ContainerImage
```
maysunfaisal, line 97: Is `outputImage` the diff between `paramsForInitialBuild` & `paramsForWebhookBasedBuilds` funcs? why is it so? 🤔

controllers/build.go
```diff
@@ -0,0 +1,263 @@
...
+				},
+			},
+			VolumeMode: &fsMode,
+		},
+	}
+	return *workspaceStorage
+}
+
+func route(component appstudiov1alpha1.Component) routev1.Route {
+	var port int32 = 8080
```
maysunfaisal, line 177: should be?
```suggestion
	var port int32 = int32(component.Spec.TargetPort)
```

controllers/component_controller.go
```diff
@@ -291,8 +298,137 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+	if err != nil {
+		log.Error(err, fmt.Sprintf("Unable to get inital webhook %v", webhook.Name))
+		return ctrl.Result{}, err
+	}
+
+	/*
+		if createdWebhook != nil && len(createdWebhook.Status.Ingress) != 0 {
+			return createdWebhook.Status.Ingress[0].Host, err
+		}
+	*/
```
maysunfaisal, line 422: comment found

controllers/build.go
```diff
@@ -0,0 +1,263 @@
...
+			Name:        component.Name,
+			Namespace:   component.Namespace,
+			Annotations: commonAnnotations(),
+		},
+		Spec: routev1.RouteSpec{
+			Path: "/",
+			To: routev1.RouteTargetReference{
+				Kind: "Service",
+				Name: "el-" + component.Name,
+			},
```
maysunfaisal, line 189: not sure i understand how this Service is getting created 🤔

controllers/component_controller.go
```diff
@@ -291,8 +298,137 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+				log.Error(err, fmt.Sprintf("Unable to create eventListener %v", eventListener))
+				return ctrl.Result{}, err
+			}
+		} else {
+			log.Error(err, fmt.Sprintf("Unable to get eventListener %v", eventListener))
+			return ctrl.Result{}, err
+		}
+	}
+
+	initialBuildSpec := determineBuildExecution(component, paramsForInitialBuild(component), "initialbuildpath")
```
maysunfaisal, line 366: could you pls help me understand the purpose of this intial build, how is it different to what was already generated during triggerTemplate above?

controllers/component_controller.go
```diff
@@ -291,8 +298,137 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		}
 	}
 
+	if component.Spec.Build.ContainerImage == "" {
```
maysunfaisal, line 301: This entire logic from here until EOF should be enclosed in 

```
if component.Spec.Source.GitSource != nil && component.Spec.Source.GitSource.URL != "" {
		// logic for Trigger Template, Event Listener and Webhook Route
}
```

controllers/build.go
```diff
@@ -0,0 +1,263 @@
...
+// determineBuildExecution returns the pipelineRun spec
+func determineBuildExecution(component appstudiov1alpha1.Component, params []tektonapi.Param, workspaceSubPath string) tektonapi.PipelineRunSpec {
+
+	pipelineRunSpec := tektonapi.PipelineRunSpec{
+		Params: params, //paramsForWebhookBasedBuilds(component),
+		PipelineRef: &tektonapi.PipelineRef{
+
+			// This can't be hardcoded to devfile-build.
+			// The logic should determine if it is a
+			// nodejs build, java build, dockerfile build or a devfile build.
```
maysunfaisal, line 53: wondering how would this happen? Is it something that HAC would know and the component.Spec should have a field to indicate the type of build?

controllers/build.go
sbose78: This is not the component's port, rather the one associated with the tekton eventlistener's service.

controllers/build.go
sbose78: It gets created upon creation of an EventListener resource.

controllers/component_controller.go
sbose78: The initial build is where a PipelineRun is kicked off right after the Component is created.
The subsequent ones are from github webhook triggers. 

The shape of the PipelineRun remains the same though.

sbose78:
Tests are 🟢  , waiting for the latest coverage report to be updated.

main.go
```diff
@@ -83,6 +87,20 @@ func main() {
 		os.Exit(1)
 	}
 
+	if err := triggersapi.AddToScheme(mgr.GetScheme()); err != nil {
+		setupLog.Error(err, "unable to add triggers api to the schema")
```
johnmcollier, line 91: Nit: 
```suggestion
		setupLog.Error(err, "unable to add triggers api to the scheme")
```

main.go
```diff
@@ -83,6 +87,20 @@ func main() {
...
+		os.Exit(1)
+	}
+
+	if err := taskrunapi.AddToScheme(mgr.GetScheme()); err != nil {
+		setupLog.Error(err, "unable to add triggers api to the schema")
```
johnmcollier, line 96: Nit:
```suggestion
		setupLog.Error(err, "unable to add triggers api to the scheme")
```

main.go
```diff
@@ -83,6 +87,20 @@ func main() {
...
+		os.Exit(1)
+	}
+
+	if err := routev1.AddToScheme(mgr.GetScheme()); err != nil {
+		setupLog.Error(err, "unable to add triggers api to the schema")
```
johnmcollier, line 101: Nit:
```suggestion
		setupLog.Error(err, "unable to add triggers api to the scheme")
```

main.go
```diff
@@ -83,6 +87,20 @@ func main() {
 		os.Exit(1)
 	}
 
+	if err := triggersapi.AddToScheme(mgr.GetScheme()); err != nil {
```
johnmcollier, line 90: Just some thoughts here, not really a review:

So this introduces a dependency on Tekton and OpenShift or HAS. I don't think we can necessarily avoid it this time (for obvious reasons, as they're needed by these changes), but being able to dev/test HAS against something considerably lighter-weight than CRC (e.g. kcp, k3s, kind, minikube) has been nice.

Wonder if [microshift](https://github.com/redhat-et/microshift) + tekton would meet the requirements for what I'm looking for in local dev? Or if there's some way we could mock routes + tekton on something like kcp/kind (like just install the CRD, but not the corresponding controller?)

controllers/component_controller.go
johnmcollier: Could we actually put the entire build logic in a separate function and call it from the reconcile function? My thinking is that the main reconcile loop for this function is already pretty long, and I'm wondering if it would help readability if we put the build functionality into its own function. 

Might also help with unit testing some of the edge cases in here, since testenv can be pretty inflexible in that regard

johnmcollier (changes requested):
So, with these changes, what's needed now to develop with HAS on a given cluster?

Could you update the README of the repository to cover all of the new and necessary prerquisites to set up HAS for dev/testing (e.g. installing third-party CRDs, setting up registry secret, etc).

It would also be nice to make the default quay.io org we push to configurable via an env var (quay.io/redhat-appstudio). Then for dev/testing I could just set something like `QUAY_ORG` and have it use my account/org by default.

sbose78:
Thanks a lot for taking out time to review!

> Additionally, with these changes, what's needed now to develop with HAS on a given cluster?

Sure, I will update the readme 👍 


> Left a couple miscellaneous comments.

Additionally, with these changes, what's needed now to develop with HAS on a given cluster?

> It would also be nice to make the default quay.io org we push to (quay.io/redhat-appstudio) configurable via an env var . Then for dev/testing I could just set something like QUAY_ORG and have it use my account/org by default. Similar to how HAS configures the GitHub org it pushes to.

Good idea, I'll make it an env var. 

Do note, that for developing HAS locally, the only other thing that you'd need would be the installation of OpenShift pipelines or the Tekton CRDs. The missing registry secret or an incorrect Quay org wouldn't impact general development - except that the  execution of the PipelineRun will fail. However, all resources should be created fine.

controllers/component_controller.go
johnmcollier: I'm wondering too, in addition to extracting this out to a function as I mentioned, I wonder if we could move the the build logic up into the end of the if / else block up above? Basically just above [here](https://github.com/redhat-appstudio/application-service/blob/26c102883a080a099d6df30840844f6999d8350d/controllers/component_controller.go#L233) for Component create, and just above [here](https://github.com/redhat-appstudio/application-service/blob/26c102883a080a099d6df30840844f6999d8350d/controllers/component_controller.go#L295) for component update.

It would also enable us to set the proper status condition (e.g. Create/Update) upon success / failures (by calling r.SetCreateConditionAndUpdateCR` and `r.SetUpdateConditionAndUpdateCR`)

During Component cr

sbose78:
✅  Make the default quay.io org we push to (quay.io/redhat-appstudio) configurable via an env var 
✅  Could we actually put the entire build logic in a separate function and call it from the reconcile function 
✅  Could you update the README of the repository to cover all of the new and necessary prerquisites to set up HAS for dev/testing (e.g. installing third-party CRDs, setting up registry secret, etc).

controllers/build.go
```diff
@@ -0,0 +1,280 @@
...
+
+// triggerTemplate generates the TriggerTemplate resources
+// which defines how a webhook-based trigger event would be handled -
+// In this case, a PipelineRun to build an image would be created.
+func triggerTemplate(component appstudiov1alpha1.Component) (*triggersapi.TriggerTemplate, error) {
+	webhookBasedBuildTemplate := determineBuildExecution(component, paramsForWebhookBasedBuilds(component), "$(tt.params.git-revision)")
+
+	resoureTemplatePipelineRun := tektonapi.PipelineRun{
+		ObjectMeta: v1.ObjectMeta{
+			GenerateName: component.Name + "-",
```
johnmcollier, line 217: So I tried out your changes, and was able to get a container build working and pushed up to quay.io, which was cool to see.

But, I saw that the pipelinerun had the name `component-sample` (the name of my Component), and stuck around _after_ the build finished or failed, and also stuck around after I deleted the Component and Application CRs.

The other thing I saw was that it was the only pipelinerun that would get created. E.g. if I changed the Component resource to something different, no other pipelineruns would get created. It was like the `GenerateName` field wasn't working for me?

controllers/application_controller_test.go
```diff
@@ -92,6 +92,51 @@ var _ = Describe("Application controller", func() {
 		})
 	})
 
+	Context("Create Application with no repositories set", func() {
```
johnmcollier, line 95: This test seems like it's duplicated?

controllers/component_controller.go
```diff
@@ -293,8 +301,145 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		component.Spec.Build.ContainerImage = r.ImageRepository + ":" + component.Namespace + "-" + component.Name
+	}
+
+	_, err = r.generateBuild(ctx, &component)
```
johnmcollier, line 308: Thanks for moving this into a function 👍 .

Could you also move it to the spots I mentioned in https://github.com/redhat-appstudio/application-service/pull/22/files#r793094792?

That'll allow us to properly set the Success or Error status conditions based on the error return value of the function.

E.g. if `generateBuild` fails right now, the status condition will be either a `Created` with a value of `True` or `Updated` with a value of `True`, when the value should actually be `False`.

controllers/component_controller.go
```diff
@@ -293,8 +301,145 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+	if err != nil {
+		log.Error(err, "Unable to update Component's status")
+	}
+
 	log.Info(fmt.Sprintf("Finished reconcile loop for %v", req.NamespacedName))
-	return ctrl.Result{}, nil
+	return ctrl.Result{}, err
+}
+
+func (r *ComponentReconciler) generateBuild(ctx context.Context, component *appstudiov1alpha1.Component) (ctrl.Result, error) {
```
johnmcollier, line 322: If we're not using `ctrl.Result` as a return value, can we just drop it from the function header?

johnmcollier (changes requested):

controllers/build.go
sbose78: That's a known issue I'd say.

The build objects are not being updated upon changes to Components ( unless they are absent/deleted ). I'll be looking into that in the upcoming PRs - especially because the next step is to programmatically persist the tekton resources in git.

sbose78: > The other thing I saw was that it was the only pipelinerun that would get created. E.g. if I updated the Component resource, no other pipelineruns would get created. It was like the GenerateName field wasn't working for me?

I remember why exactly I didn't use the `generateName` for PipelineRun - I was getting too many PipelineRuns since there are usually multiple controller reconciles from the same update sometimes. I'd like to fix that.

> But, I saw that the pipelinerun had the name component-sample (the name of my Component), and stuck around after the build finished or failed, and also stuck around after I deleted the Component and Application CRs.

This is to ensure that the record of the build isn't lost upon deletion of a `Component`. Cleaning up of PipelineRuns would be taken care of a parallel pruning activity ( which isn't in place yet ).

controllers/application_controller_test.go
sbose78: Fixed it - was a duplicate name.

controllers/build.go
sbose78: I take that back, it makes sense to delete PipelineRuns too upon deletion of a Component, that is the only way to clean the namespace up after usage. 

Let me fix that.

controllers/build.go
```diff
@@ -0,0 +1,280 @@
...
+			Namespace:    component.Namespace,
```
johnmcollier, line 218: Just realizing something here. So since this PipelineRun will run in the namespace the component has been created in, and since components can be created in any namespace, this means each namespace will need to have the `redhat-appstudio-registry-pull-secret` secret created, right?

Probably not an issue for Milestone 2, but something to think about. @sbose78 wdyt?

sbose78: Correct. 
This is why we granted create & delete permissions to users for `secrets` in app studio namespaces.

However, it's a painful user experience because the UI doesn't have a way to take the input image (yet)- therefore the user will need to be able to push to the pre-defined image repository for her token to work. It gets easier if done from the CLI - since the user can specify both the image repo ( .spec.build.containerImage ) and create the `secret`

Given that, users don't have a way to view the contents of secrets in their namespaces, I'm thinking of having the  `redhat-appstudio-registry-pull-secret` created in every namespace at the time of the component creation ( same way the `appstudio` pvc is being setup now per namespace in this PR ). They can of course delete it and create their own secret if they wish to. What do you think ?

Mid-term/Long-term, I'd consider setting up of the namespace/workspace specific assets at the time of workspace/namespace provisioning - and not at the time of Component creation.

sbose78:
( I will have to test this on a cluster after moving things around )

controllers/build.go
johnmcollier: 👍 for both suggestions.

controllers/build.go
```diff
@@ -0,0 +1,282 @@
...
+}
+
+// route returns the Route resource that would enable
+// ingress traffic into the webhook endpoint ( aka EventListener)
+// TODO: This needs to be secure.
+func route(component appstudiov1alpha1.Component) routev1.Route {
+	var port int32 = 8080
+	webhook := &routev1.Route{
+		ObjectMeta: v1.ObjectMeta{
+			Name:        component.Name,
```
johnmcollier, line 193: Should we use el-" + component.Name for the route name here? We also use the component name for the route that we generate in the gitops repository (see https://github.com/redhat-appstudio/application-service/blob/main/gitops/generate.go#L170), so there would be a conflict between these two routes

johnmcollier (commented):
Changes generally look good to me, and work for me on CRC. 

Just one last concern regarding the el route name.

maysunfaisal:
@sbose78 I am also encountering an intermittent error:

```
2022-01-28T20:22:25.357Z	INFO	controllers.Application	Finished reconcile loop for application-service-system/application-sample	{"Application": "application-service-system/application-sample"}
2022-01-28T20:22:25.358Z	INFO	controllers.Component	Updating the labels for Component application-service-system/component-sample	{"Component": "application-service-system/component-sample"}
2022-01-28T20:22:26.752Z	ERROR	controllers.Component	Unable to get inital webhook component-sample	{"Component": "application-service-system", "error": "Route.route.openshift.io \"component-sample\" not found"}
github.com/redhat-appstudio/application-service/controllers.(*ComponentReconciler).Reconcile
	/workspace/controllers/component_controller.go:240
```

it seems like after the Route is created, we immediately try to get it and read it. We dont do a wait after creation and this maybe when the err occurs? because when I get routes its there:

```
MacBook-Pro:application-service maysun$ oc get routes
NAME               HOST/PORT                                                                                                PATH   SERVICES              PORT   TERMINATION   WILDCARD
component-sample   component-sample-application-service-system.apps.ci-ln-ff0h25t-76ef8.origin-ci-int-aws.dev.rhcloud.com   /      el-component-sample   8080                 None
```

controllers/build.go
sbose78: @maysunfaisal Im wondering if ComponentDetectionQuery would help us with those?

controllers/build.go
sbose78: Good point, let me rename that..

sbose78:
> it seems like after the Route is created, we immediately try to get it and read it. We dont do a wait after creation and this maybe when the err occurs? because when I get routes its there:

Thanks for reporting this!
That's a definite possibility, let me what I can do.

sbose78:
@maysunfaisal Did the reconcile loop error reporting re-reconcile and eventually stabilize ?

controllers/build.go
sbose78: Done f3d6037

controllers/component_controller.go
sbose78: It's a true `generateName` now

https://github.com/redhat-appstudio/application-service/pull/22/files#diff-9f23e6f71fd9e4b8a6dc2c3cda6f3dc30f20ba7519030cbbf46d5085e0dfcdfaR413

controllers/build.go
sbose78: If a Component is updated,
* a new PipelineRun is created ( it uses a generateName now )
* the existing TriggerTemplate is updated

main.go
sbose78: ( getting back to this to share my thoughts shortly )

mmorhun:
Just retested permissions part - now everything works fine for me.

johnmcollier (approved):
LGTM

sbose78:
Thank you!

As a general project backlog item, we'll need to set up common code infra for this repo to mock the api server to be able to test the error paths better. That is the key reason why the code coverage is plunging right now.

PR #29, opened 2022-01-13
[DEVHAS-61] Allow the GitOps org to be configured easily

johnmcollier:
Fixes https://issues.redhat.com/browse/DEVHAS-61

This PR allows HAS to be easily configured to point to an alternate GitHub org by allowing a ConfigMap to be created before deployment.

In all:
- Removes the github.properties file from the deployment manifests for HAS
  - It's not needed and actually makes it harder to configure the GitHub org that HAS uses
- No longer generates a ConfigMap from this github.properties file
  - But the optional ConfigMap mount is still kept in the operator's deployment
- Updates the readme to reflect these changes


johnmcollier:
We might not need this with https://github.com/redhat-appstudio/infra-deployments/pull/58

/hold

PR #34, opened 2022-01-21
Clean up HAS and improve test coverage

johnmcollier:
- Removes the Push CRD from HAS
- Some of the Operator's configuration yamls still had the has prefix in their filename, so I renamed them to drop the prefix
- Adds some additional test cases in the following areas to improve test coverage
   - Application CRD Finalizer
   - Component Controller
   - ComponentDetectionQuery
   - GitOps packages
   - Util package


gitops/yaml/resources.go
```diff
@@ -75,10 +75,3 @@ func MarshalOutput(out io.Writer, output interface{}) error {
 	}
 	return nil
 }
-
-// AddKustomize adds kustomization.yaml.  Name and items become map key and value, respectively
-func AddKustomize(fs afero.Fs, name string, items []string, path string) error {
```
johnmcollier, line 80: Duplicated function for Kustomize that was never used.

maysunfaisal (approved):

PR #36, opened 2022-01-25
Watch secret with Github token

AndrienkoAleksandr:
Motivation: gh token can be revoked or expired. It would be nice to allow update this token via secret without operator deployment rescale.

### DIFF
--- a/controllers/application_controller.go
+++ b/controllers/application_controller.go
@@ -65,15 +65,22 @@ type ApplicationReconciler struct {
 func (r *ApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	log := r.Log.WithValues("Application", req.NamespacedName)
 
-	if req.Name == githubTokenSecret || r.Context.GithubConf.GithubClient == nil {
-		githubTokenSecret := &corev1.Secret{}
-		if err := r.Client.Get(ctx, types.NamespacedName{Namespace: req.Namespace, Name: req.Name}, githubTokenSecret); err != nil {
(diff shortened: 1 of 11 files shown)


controllers/application_controller.go
```diff
@@ -205,7 +229,20 @@ func (r *ApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 func (r *ApplicationReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	gofakeit.New(0)
 
+	mapSecretsFunc := func(obj client.Object) []reconcile.Request {
```
johnmcollier, line 232: So, does this mean HAS would only watch secrets with the name `has-github-token`, and no other secrets would be watched? What if `has-github-token` doesn't exist? Will it still work fine? (e.g. if we're running the HAS binary locally with the `GITHUB_AUTH_TOKEN` env var exported)

johnmcollier (changes requested):
Left a couple comments related to testing.

controllers/application_controller.go
AndrienkoAleksandr: > So, does this mean HAS would only watch secrets with the name has-github-token, and no other secrets would be watched?

We can extends this function any time to watch more secrets. For now application service uses only one secret to store some part of the configuration.

> What if has-github-token doesn't exist? Will it still work fine? (e.g. if we're running the HAS binary locally with the GITHUB_AUTH_TOKEN env var exported)

Yes, it will work fine with env variable.

johnmcollier:
@AndrienkoAleksandr Is this PR still needed? Or can it be closed?

PR #37, opened 2022-01-26
Treat URLs with .git as valid

johnmcollier:
Fixes https://issues.redhat.com/browse/DEVHAS-63

Updates the `util.ConvertGitHubURL` function to treat GitHub URLs with `.git` at the end as valid.

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -51,8 +51,15 @@ func IsExist(path string) (bool, error) {
 }
 
 // ConvertGitHubURL converts a git url to its raw format
-// taken from Jingfu's odo code
+// adapted from https://github.com/redhat-developer/odo/blob/e63773cc156ade6174a533535cbaa0c79506ffdb/pkg/catalog/catalog.go#L72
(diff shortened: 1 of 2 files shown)


pkg/util/util.go
```diff
@@ -51,8 +52,16 @@ func IsExist(path string) (bool, error) {
...
 
 // ConvertGitHubURL converts a git url to its raw format
-// taken from Jingfu's odo code
+// adapted from https://github.com/redhat-developer/odo/blob/e63773cc156ade6174a533535cbaa0c79506ffdb/pkg/catalog/catalog.go#L72
 func ConvertGitHubURL(URL string) (string, error) {
+	// If the URL ends with .git, remove it
+	// The regex will match if '.git' is at the end of the given string
+	reg, err := regexp.Compile(".git$")
+	if err != nil {
+		return "", err
```
johnmcollier, line 61: Since we're using a hardcoded regexp here, there isn't really a way to unit test this error 😕

maysunfaisal (approved):

PR #40, opened 2022-02-01
Add a simple retry logic for fetching the route after creation

sbose78:
Tested this change, and 100% of the time the route information could be fetched. Hope it actually works ;) Either way, this change stops reporting the error because it's clearly a false positive. Worst case, the `status.webhook` url shows up in the next reconcile.

I'll follow up with something more sophisticated.

### DIFF
--- a/controllers/component_controller.go
+++ b/controllers/component_controller.go
@@ -471,7 +471,7 @@ func (r *ComponentReconciler) generateBuild(ctx context.Context, component *apps
 		}
 
 		if createdWebhook != nil && len(createdWebhook.Status.Ingress) != 0 {
-			log.Info(fmt.Sprintf("Github webhook url generated %v", createdWebhook.Status.Ingress[0].Host))
+			log.Info(fmt.Sprintf("webhook url generated %v", createdWebhook.Status.Ingress[0].Host))
 			return createdWebhook.Status.Ingress[0].Host, nil
 		}
 	}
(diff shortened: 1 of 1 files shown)


sbose78:
No new tests because : 

(1) the logic hasn't change, I've only moved it inside a loop of 3 retries.
(2) there's no infra for api server mocking yet - once we have that, we could try having a 'delay' test.

We already have tests for looking up the created route.

mmorhun (approved):
I think we should improve it in the future

controllers/component_controller.go
```diff
@@ -455,18 +456,24 @@ func (r *ComponentReconciler) generateBuild(ctx context.Context, component *apps
...
-		webhookURL := createdWebhook.Status.Ingress[0].Host
-		log.Info(fmt.Sprintf("Github webhook url generated %v", webhookURL))
-		return webhookURL, err
+	for i := 0; i < 3; i++ {
+		time.Sleep(time.Second * 1)
+		// Ideally, one must wait for the route to be 'accepted'?
+		createdWebhook := &routev1.Route{}
+		err := r.Client.Get(ctx, types.NamespacedName{Name: webhook.Name, Namespace: webhook.Namespace}, createdWebhook)
+		if err != nil {
+			log.Error(err, fmt.Sprintf("Unable to fetch the created webhook %v, retrying", webhook.Name))
```
sbose78, line 470: Silently failing on purpose because not being able to read the `route` isn't a show-stopper.

maysunfaisal (approved):

PR #42, opened 2022-02-02
Add build gitops generation

mmorhun:

#### What this MR does
This PR adds functionality to create build related gitops resources under `.tekton` folder.

#### Which issue this MR fixes
https://issues.redhat.com/browse/PLNSRVCE-46

#### How to test
1. Create an application and a component for it:
```yaml
kind: Application
metadata:
  name: application-sample
  namespace: test
spec:
  description: test simplae petclinic app
  displayName: test-petclinic
```
```yaml
kind: Component
metadata:
  name: component-sample
  namespace: test
spec:
  source:
    git:
      url: https://github.com/devfile-samples/devfile-sample-code-with-quarkus
  build:
    containerImage: docker.io/<your-account>/foobar:dev-code-with-quarkus

```
2. Check your GitHub organization: it must contain a dedicated gitops repository with the component folder that includes main gitops for the components and build gitops under `.tekton` folder in the component's directory:
Screenshot from 2022-02-02 12-44-11
Screenshot from 2022-02-02 12-44-32

### DIFF
--- a/controllers/component_controller.go
+++ b/controllers/component_controller.go
@@ -230,7 +230,7 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 
 				// Generate and push the gitops resources if spec.containerImage is set
 				if component.Spec.Build.ContainerImage != "" {
-					err = r.generateGitops(&component)
+					err = r.generateGitops(&component, gitops.Generate)
 					if err != nil {
 						errMsg := fmt.Sprintf("Unable to generate gitops resources for component %v", req.NamespacedName)
 						log.Error(err, errMsg)
@@ -298,7 +298,7 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 			// Generate and push the gitops resources if spec.containerImage is set
 			if component.Spec.Build.ContainerImage != "" {
 				component.Status.ContainerImage = component.Spec.Build.ContainerImage
-				err = r.generateGitops(&component)
+				err = r.generateGitops(&component, gitops.Generate)
 				if err != nil {
 					errMsg := fmt.Sprintf("Unable to generate gitops resources for component %v", req.NamespacedName)
 					log.Error(err, errMsg)
@@ -333,7 +333,9 @@ func (r *ComponentReconciler) generateBuild(ctx context.Context, component *apps
 
 	log := r.Log.WithValues("Namespace", component.Namespace, "Application", component.Spec.Application, "Component", component.Name)
 
-	workspaceStorage := commonStorage(*component, "appstudio")
+	// TODO delete creation of gitops build objects(except PipelineRun) when build part of gitops repository will be respected
+
+	workspaceStorage := gitops.GenerateCommonStorage(*component, "appstudio")
 	pvc := &corev1.PersistentVolumeClaim{}
 	err := r.Get(ctx, types.NamespacedName{Name: workspaceStorage.Name, Namespace: workspaceStorage.Namespace}, pvc)
 	if err != nil {
@@ -350,7 +352,7 @@ func (r *ComponentReconciler) generateBuild(ctx context.Context, component *apps
 	}
 	log.Info(fmt.Sprintf("PV is now present : %v", workspaceStorage.Name))
 
-	triggerTemplate, err := triggerTemplate(*component)
+	triggerTemplate, err := gitops.GenerateTriggerTemplate(*component)
 	if err != nil {
 		log.Error(err, "Unable to generate triggerTemplate ")
 		return "", err
@@ -384,7 +386,7 @@ func (r *ComponentReconciler) generateBuild(ctx context.Context, component *apps
 		}
 		log.Info(fmt.Sprintf("TriggerTemplate updated %v", triggerTemplate.Name))
 	}
-	eventListener := eventListener(*component, *triggerTemplate)
+	eventListener := gitops.GenerateEventListener(*component, *triggerTemplate)
 
 	err = controllerutil.SetOwnerReference(component, &eventListener, r.Scheme)
 	if err != nil {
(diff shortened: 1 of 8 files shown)


sbose78:
Please chat with @elsony , @johnmcollier and @maysunfaisal to convey in details what/how this is being done :)

controllers/component_controller.go
```diff
@@ -230,7 +230,7 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 
 				// Generate and push the gitops resources if spec.containerImage is set
 				if component.Spec.Build.ContainerImage != "" {
```
mmorhun, line 232: Is it possible to have `component.Spec.Build.ContainerImage` empty? Seems it's set to default value [here](https://github.com/redhat-appstudio/application-service/blob/46684055dc75b2e05899d8d38cf209de17460ace/controllers/component_controller.go#L97).

maysunfaisal: yeah true, it does seem the default is set there. It was set via a different PR so this check was not removed.

mmorhun: Obsolete conditions removed.

Michkov:
Created this repository - https://github.com/Michkov-test/petclinic-default-invite-return

1. components/component-sample/base/kustomization.yaml -> probably should include `.tekton`
2. [.tekton](https://github.com/Michkov-test/petclinic-default-invite-return/tree/main/components/component-sample/base/.tekton) all files are missing root `kind` and `apiVersion`

gitops/generate_build.go
```diff
@@ -14,36 +14,75 @@ See the License for the specific language governing permissions and
...
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
-func determineBuildDefinition(component appstudiov1alpha1.Component) string {
-	return "devfile-build"
-}
+const (
+	buildCommonStoragePVCFileName = "common-storage-pvc.yaml"
+	buildTriggerTemplateFileName  = "trigger-template.yaml"
+	buildEvetListenerFileName     = "event-listener.yaml"
```
maysunfaisal, line 41:
```suggestion
	buildEventListenerFileName     = "event-listener.yaml"
```

gitops/generate_build.go
```diff
@@ -14,36 +14,75 @@ See the License for the specific language governing permissions and
...
+	buildWebhookRouteFileName     = "build-webhook-route.yaml"
+)
 
-func determineBuildCatalog(namespace string) string {
-	// TODO: If there's a namespace/workspace specific catalog, we got
-	// to respect that.
-	return "quay.io/redhat-appstudio/build-templates-bundle@sha256:2205a29208fa686b47f841819f7abedb64adb93935493693892d0e18bbdbb77e"
+func GenerateBuild(fs afero.Fs, outputFolder string, component appstudiov1alpha1.Component) error {
+	// Because GenerateAndPush cleans up everything for component it cleans previously generated main gitops files for the component.
```
maysunfaisal, line 46: I wonder if we should still do it once though, maybe we can streamline the basic kube and gitops resource generation in one function and call it strategically once in the controller..

Is there a case where we want the basic deployment resource generated but not the gitops resources? If not, then should we pursue it?

gitops/generate_build.go
```diff
@@ -14,36 +14,75 @@ See the License for the specific language governing permissions and
...
+		buildEvetListenerFileName:     eventListener,
+		buildWebhookRouteFileName:     webhookRoute,
+	}
+
+	kustomize := resources.Kustomization{}
+	for fileName := range buildResources {
+		kustomize.AddResources(fileName)
+	}
+
+	buildResources[kastomizeFileName] = kustomize
```
maysunfaisal, line 77:
```suggestion
	buildResources[kustomizeFileName] = kustomize
```

maysunfaisal (commented):
some initial comments

gitops/generate_build.go
mmorhun: I was thinking about it. Talked with @sbose78 and he adviced to keep build resources in a separate commit, so if for some reason generation of build gitops resources fails, it won't break main ones.

mmorhun:
@Michkov your remarks are fixed.

gitops/generate_build.go
sbose78: @maysunfaisal @mmorhun I wanted to keep it separate so that the failure of one doesn't impact the other. However, I don't have any other strong reasons for that. I'm open to doing things differently if one of you have strong feelings either way :)

mmorhun: A way with two commits is already implemented in this PR

mmorhun: But I still have no strong opinion on this

gitops/yaml/resources.go
```diff
@@ -82,3 +85,30 @@ func AddKustomize(fs afero.Fs, name string, items []string, path string) error {
 	content = append(content, map[string]interface{}{name: items})
 	return MarshalItemToFile(fs, path, content)
 }
+
+// MergeKustomizeResource rewrites existing kustomization.yaml by adding specified items under resources section
+func MergeResourcesIntoKustomize(fs afero.Fs, name string, res ...string) error {
```
johnmcollier, line 90: Can you add a test for this function?

https://github.com/redhat-appstudio/application-service/blob/main/gitops/yaml/resources_test.go should provide some pointers on how we unit tested some of the fs operations with afero (including read only fs)

gitops/generate_build.go
johnmcollier: Is there a reason why the generation of the build resources would fail that we would want to keep it separate? If it's not likely, I'm inclined to agree with Maysun and would favour doing it all at once.

johnmcollier (changes requested):

sbose78:
I'm good with having it all in one commit 👍

mmorhun:
Ok, will provide another PR with the solution that uses only one commit

PR #46, opened 2022-02-03
Don't sleep if EventListener route can't be found

johnmcollier:
I saw that the runtime of the tests had nearly doubled from ~35s to ~65s. Doing some debugging, I saw that it was due to us [sleeping while we wait for the Route resource to appear](https://github.com/redhat-appstudio/application-service/blob/main/controllers/component_controller.go#L454).

Instead of sleeping, I updated the code that gets the route to return and requeue itself if the route isn't found. This is an approach we've used in the past with the [devfile registry operator](https://github.com/devfile/registry-operator/blob/main/controllers/devfileregistry_controller.go#L119), as it seems to help refresh the cached Kube client faster, allowing it to pick up the created route quicker than sleeping

I also updated the tests for these changes. To do so, I manually update the route's status to set the ingress url, then I trigger a reconcile on the corresponding Component resource to see if its `status.webhook` field gets set to the route URL.
  - As a result, I had to update the mock Route CRD that we use in the tests to mark the status field as a subresource, otherwise `k8sclient.Status().Update()` would fail
  
@sbose78, what do you think of these changes?

### DIFF
--- a/controllers/component_controller.go
+++ b/controllers/component_controller.go
@@ -506,7 +506,7 @@ func (r *ComponentReconciler) generateGitops(component *appstudiov1alpha1.Compon
 	return r.AppFS.RemoveAll(tempDir)
 }
 
-// setGitopsStatus adds the necessary gitops annotations (url, branch, context) to the component CR status
+// setGitopsStatus adds the necessary gitops info (url, branch, context) to the component CR status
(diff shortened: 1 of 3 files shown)


sbose78 (approved):
Interesting approach, let's roll with this. Thanks!

PR #54, opened 2022-02-15
Setup CI using 'AppStudio Staging CI' based on Tekton

sbose78:
To be used with https://github.com/redhat-appstudio/infra-deployments/pull/115

### DIFF
--- a/.tekton/pull-request.yaml
+++ b/.tekton/pull-request.yaml
@@ -1,0 +1,73 @@
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  generateName: on-pr-
+  namespace: application-service
+  annotations:
+    pipelinesascode.tekton.dev/on-event: "[pull_request]"
+    pipelinesascode.tekton.dev/on-target-branch: "[main]"
+    pipelinesascode.tekton.dev/max-keep-runs: "2"
+spec:
+  params:
+    - name: repo_url
+      value: 'https://github.com/sbose78/application-service'
+    - name: revision
+      value: 'main'
+  pipelineSpec:
+    params:
+      - name: repo_url
+        type: string
+      - name: revision
+        type: string
+    tasks:
+      - name: fetchit
+        params:
+          - name: url
+            value: $(params.repo_url)
+          - name: revision
+            value: $(params.revision)
+        taskRef:
+          kind: ClusterTask
+          name: git-clone
+        workspaces:
+          - name: output
+            workspace: source
+      - name: build
+        runAfter:
+          - fetchit
+        taskSpec:
+          metadata: {}
+          steps:
+            - image: >-
+                registry.access.redhat.com/ubi8/go-toolset
+              name: build
+              resources: {}
+              script: |
+                #!/usr/bin/env bash
+                set -eux
+
+                ls -ltr 
+
+                make build
+              workingDir: $(workspaces.source.path)
+          workspaces:
+            - name: source
+        workspaces:
+          - name: source
+            workspace: source
+    workspaces:
+      - name: source
+  serviceAccountName: pipeline
+  timeout: 1h0m0s
+  workspaces:
+    - name: source
+      volumeClaimTemplate:
+        metadata:
+          creationTimestamp: null
+        spec:
+          accessModes:
+            - ReadWriteOnce
+          resources:
+            requests:
+              storage: 3Gi
+        status: {}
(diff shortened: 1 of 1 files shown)


.tekton/pull-request.yaml
```diff
@@ -0,0 +1,73 @@
...
+      volumeClaimTemplate:
+        metadata:
+          creationTimestamp: null
+        spec:
+          accessModes:
+            - ReadWriteOnce
+          resources:
+            requests:
+              storage: 1Gi
+        status: {}
```
chmouel, line 73: nit:  can be cleaneup

.tekton/pull-request.yaml
```diff
@@ -0,0 +1,73 @@
...
+                registry.access.redhat.com/ubi8/go-toolset
+              name: build
+              resources: {}
+              script: |
+                #!/usr/bin/env bash
+                set -eux
+
+                ls -ltr 
+
+                make build
```
johnmcollier, line 51: Should we either:

1) Build the operator docker image instead of the operator binary here, or
2) Or, have a separate step to build the operator docker image?

Just want to make sure that any changes to the operator dockerfile also get included in PR builds 🙂

johnmcollier: Couple things I realized re: testing.

1) Can you add a step for running `make test` after building the Operator?
2) Will this replace our current [CI GitHub Action](https://github.com/redhat-appstudio/application-service/blob/main/.github/workflows/pr.yml)? If so, how will we manage uploads to Codecov?

chmouel: fyi:

you can do both (have the CI GH action managing codecov) and other stuff with pac, if you want to go full black you can snatch this code  : 

https://github.com/openshift-pipelines/pipelines-as-code/blob/main/.tekton/pull-request.yaml#L147

you just need to make sure to add the secret in the namespace with the codecov token

sbose78:
> Will this run on the App Studio staging cluster? If so, what roles are needed to view the CI logs? view on application-service?

Yes.

.tekton/pull-request.yaml
sbose78: > Have a separate step to build the operator docker image?

Yes, I'd actually go for a step to build the image, tag it appropriately and push it, so that one can test it by deploying it if needed.

For now, I kept it to `make build` to be able to verify that the entire works end to end :)

sbose78:
> Will this run on the App Studio staging cluster?

I'll need to add a Repository CR in the `application-service` namespace so that PipelineRuns get created there.

.tekton/pull-request.yaml
```diff
@@ -0,0 +1,71 @@
...
+            - name: source
+        workspaces:
+          - name: source
+            workspace: source
+    workspaces:
+      - name: source
+  serviceAccountName: pipeline
+  workspaces:
+    - name: source
+      volumeClaimTemplate:
```
sbose78, line 63: Future PR: will replace this with https://github.com/redhat-appstudio/infra-deployments/blob/main/components/has/.tekton/pvc.yaml

sbose78:
> I'll need to add a Repository CR in the application-service namespace so that PipelineRuns get created there.

https://github.com/redhat-appstudio/infra-deployments/pull/115/files#diff-ec798dfb834a5a1748e53d815d44fd0970c6e7bf04197269f7d4f44e0f1f03e6

sbose78:
@johnmcollier Given this is harmless, could we take this PR in soon? I'll follow that up with  https://github.com/redhat-appstudio/infra-deployments/pull/115 and then creating the Github app `secret`.

johnmcollier (approved):
Approving since this won't impact our current CI.

.tekton/pull-request.yaml
```diff
@@ -0,0 +1,71 @@
...
+          workspaces:
+            - name: source
+        workspaces:
+          - name: source
+            workspace: source
+    workspaces:
+      - name: source
+  serviceAccountName: pipeline
+  workspaces:
+    - name: source
```
sbose78, line 62: @chmouel how do I use the equivalent of https://github.com/redhat-appstudio/infra-deployments/blob/main/components/gitops/.tekton/trigger-template.yaml#L37 here ? 

I'd like all 'git clones' to happen in a subpath `something-$(tt.params.git-revision)`. 

fyi @Michkov

chmouel: you can use pac variable interpolation here: https://github.com/openshift-pipelines/pipelines-as-code/blob/main/docs/guide.md#authoring-pipelinerun-in-tekton-directory #54

ie: 



```yaml
            subPath: gitops-service-{{revision}}
```

PR #59, opened 2022-02-16
Rename the pipelinerun yaml 

sbose78:
to avoid the error in https://github.com/redhat-appstudio/application-service/pull/58

### DIFF
--- a/.tekton/pull-request.yaml
+++ b/.tekton/pull-request.yaml
@@ -1,7 +1,7 @@
 apiVersion: tekton.dev/v1beta1
 kind: PipelineRun
 metadata:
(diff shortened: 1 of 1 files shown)


johnmcollier (approved):

.tekton/pull-request.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: tekton.dev/v1beta1
 kind: PipelineRun
 metadata:
-  generateName: on-pr-
+  name: has-on-pull-request
   namespace: application-service
```
chmouel, line 5: you don't need to specify the namespace either it will be tested where the repository is

PR #61, opened 2022-02-16
What would it take to do a `make test` 

sbose78:
.. using tekton


sbose78:
looks like it just works

PR #62, opened 2022-02-24
Build images for repos with/without devfiles

mmorhun:
The work is in progress. This is still a prototype.

#### Issues this MR fixes
https://issues.redhat.com/browse/PLNSRVCE-66


config/crd/bases/appstudio.redhat.com_components.yaml
```diff
@@ -333,6 +333,10 @@ spec:
                       component
                     type: string
                 type: object
+              pipeline:
```
sbose78, line 336: How would this be filled?

mmorhun: Name of pipeline to use for build, e.g. `devfile-build`, `docker-build`, etc.

jduimovich:
Can I assume the discussion ended up concluding we will always have a devfile to consult ?  
1. Couple points, we still will need to map from component query to the correct build pipeline
2. the devfile build pipeline currently extracts outerloop information from the devfile from the repository .  In the use case where the git repo does not included a devfile we will have to move the devfile inspection from the pipeline to the processing of the componentquery.

sbose78:
@jduimovich 

Here's how the discussion went:

We should not care at all where the `devfile` file is to be honest.

1. Look for the model in the component's .status.devfile ( could have been generated from the /devfile in repo or devfile registry )
2. If the model says that there's a Dockerfile present, use the Dockerfile builder from our tekton bundle.
3. If Dockerfile not present, use the detected runtime to determine the right pipeline from our tekton bundle.

Do you think this is a good start? :)

jduimovich:
@sbose78 @mmorhun 

>>concluding we will always have a devfile to consult

> We should not care at all where the devfile file is to be honest.

Right cause its in the `component's .status.devfile ` :) 

> Do you think this is a good start? :)

Yes, that will do. 

Essentially, all I was trying to summarize in my prior comment was some of the changes we need to make
- remove the current current devfile build as we'll ignore the devfile in the users repo
- we need to do the mapping from runtime -> pipeline for non-dockerfile cases so  As long as we get the analysis from the component we can do our mapping there. We need to make mapping configurable (not hard coded) 
- there are edge cases, in our current java examples - they have a dockerfile and a pom.xml and the dockerfile builder is the incorrect choice there since you need to build java first and then package it
- this also means we will have to re-inspect the component if it ever changes and possible re-do the pipleine choices and update any places we generated it as a hardcoded thing (triggers) 
- we should also update the test scripts to remove our emulation of the runtime detection and use the component query and component launcher to exercise the detect and launch flow. (and yes we already agreed you should not need to create a component to launch a build but  ... c'est la vie)  
- your #3 means I think if we inspect the alizer language analysis from the runtime we can work with we can also support require pre-devfile 2.2 (which is missing an outerloop section)  Maybe helps with .NET

mmorhun:
Create a new MR that implements pipeline selection: https://github.com/redhat-appstudio/application-service/pull/67

mmorhun:
> We need to make mapping configurable (not hard coded)

@jduimovich could you elaborate on how you see them configurable? (I am +1 for this, but not sure about the best way how user could do it: via a field in component CR? An annotation?)

jduimovich:
We incrementally can solve a few problems

- Allowing override of our default choice and enables fixing when we cannot find the right pipeline.   

We annotate the Component with the pipeline name that was discovered for it by the pipeline service code.  To allow override we need to detect changes to these annotations and regenerate triggers etc.

To solve the issue of the pipeline not found  I think we should change the pipeline to  return a "no-pipeline-found" instead of noop pipeline which will run and print instructions for the user on how  how to set the annotation (as above) to select the right pipeline. 

Second, the configurability.  In your MR, I saw bunches of logic for the mapping.
If we simplify,. this is easier to modify and if we decide, a configmap could be introduced to hold this mapping for external configuration override of the internal defaults.  For example, we should be able to rename all our pipeline names and have it work with no changes to go code. 
```
func languageToPipeline(language string) string {
	languageMap := map[string]string{
		"java":    "java-builder",
		"node":    "nodejs-builder",
		"default": "noop",
	}
	var pipeline = languageMap[language]
	if pipeline == "" {
		return languageMap["default"]
	}
	return pipeline
}
```

mmorhun:
@jduimovich sounds good to me.
A tiny detail: beside language detection, there is framework detection which eventually should/can be used.

jduimovich:
Yes, the framework is a good one as there will be framework specific pipelines. We also need other information exposed - for example node.js package.json includes an `engines:` tag which tells you which runtime version to use. Java has a mechanism as well, so when we update to support multiple runtimes, the model needs to expose that.

PR #63, opened 2022-03-02
[DEVHAS-67] Add support for private git repos

johnmcollier:
This PR adds support for private Git repositories to HAS, for `Component` and `ComponentDetectionQuery` resources, by using SPI and [scm-file-retriever](https://github.com/redhat-appstudio/service-provider-integration-scm-file-retriever).

Since SPI does not yet support Git clone operations, the token secret created from an SPIAccessTokenBinding resource needs to be passed in to the Component when using private Git repositories, as we need to retrieve the token from the secret.

**To use a private Git repo with a `Component` resource:**
1) Create an `SPIAccessTokenBinding` for the given Git repository
2) Create a `Component` resource for the Git repository, and specifying the Kubernetes secret created from the previous step (`Spec.Source.GitSource.Secret`)

**To use a private Git repo with a `ComponentDetectionQuery` resource:**
1) Create an `SPIAccessTokenBinding` for the given Git repository
2) Create a `ComponentDetectionQuery` resource for the Git repository, and specifying the Kubernetes secret created from the previous step (`Spec.GitSource.Secret`)


johnmcollier:
Instructions on setting up SPI w/ HAS https://github.com/redhat-appstudio/application-service/blob/c79a43f53c1c2209c271111e3e7eecb7947bdd75/docs/private-git-repos.md

go.mod
```diff
@@ -12,18 +12,137 @@ require (
...
+	k8s.io/api v0.22.5
+	k8s.io/apimachinery v0.22.5
+	k8s.io/client-go v0.22.5
+	sigs.k8s.io/controller-runtime v0.10.3
 	sigs.k8s.io/yaml v1.3.0
 
 )
+
+require (
+	cloud.google.com/go v0.98.0 // indirect
```
maysunfaisal, line 33: is this because all the other mods havent upgraded to 1.17?

controllers/component_controller.go
```diff
@@ -99,36 +101,68 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
 			if source.GitSource.DevfileURL == "" {
-				rawURL, err := util.ConvertGitHubURL(source.GitSource.URL)
-				if err != nil {
-					log.Error(err, fmt.Sprintf("Unable to convert Github URL to raw format, exiting reconcile loop %v", req.NamespacedName))
-					r.SetCreateConditionAndUpdateCR(ctx, &component, err)
-					return ctrl.Result{}, nil
-				}
 
 				// append context to the path if present
 				// context is usually set when the git repo is a multi-component repo (example - contains both frontend & backend)
```
maysunfaisal, line 133: can you move this comment to the right place, I was confused for a moment lol

controllers/component_controller.go
```diff
@@ -99,36 +101,68 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+
+					devfileBytes, err = util.DownloadDevfile(devfileDir)
+					if err != nil {
+						log.Error(err, fmt.Sprintf("Unable to read the devfile from dir %s %v", devfileDir, req.NamespacedName))
+						r.SetCreateConditionAndUpdateCR(ctx, &component, err)
+						return ctrl.Result{}, nil
+					}
 				} else {
-					devfileDir = path.Join(rawURL, context)
+					// Use SPI to retrieve the devfile from the private repository
```
maysunfaisal, line 157: Just a curiosity, so we assume its private repo based on if a token is provided. Wondering if there is a way if scm file retriever mod can figure that out for us and just use it.. so we can have file retriever for both public and private repo

config/rbac/role_spi.yaml
```diff
@@ -0,0 +1,24 @@
...
+    - appstudio.redhat.com
+  resources:
+    - spiaccesstokenbindings
+    - spiaccesstokens
+- verbs:
+    - create # allows addition of credentials only.
+    - delete
+    - list
+    - get
+    - watch
```
maysunfaisal, line 20: do we need this? I dont see it being used 🤔

pkg/util/util_test.go
```diff
@@ -208,11 +209,18 @@ func TestCloneRepo(t *testing.T) {
 			repo:      "https://github.com/devfile-samples/devfile-sample-java-springboot-basic",
 			wantErr:   false,
 		},
+		{
```
maysunfaisal, line 212: I assume there is no straightforward way to test a happy path

maysunfaisal (commented):
Just some questions, looks nice, thanks

I also just wanted to ask about the `GitSource.DevfileUR` path in both our controllers, a user can potentially also provide a raw url. Not sure how that works.. when I clicked on the raw link form the Github UI, it pasted a token as well. So I assume we should be good?

config/rbac/role_spi.yaml
johnmcollier: Needed here https://github.com/redhat-appstudio/application-service/pull/63/files#diff-3fd064f2f1027a954b035a6d5885b96cc5b483f04889d14fa840ace5e9c29c55R33

Otherwise calls to scm-file-retriever will fail with rbac errors

pkg/util/util_test.go
johnmcollier: Not without:
1) A real git token and real private repository to use
or
2) Mocking go-git's CloneRepo

2 might be worth exploring in the future, especially if we need to do anything else with go-git, but I'd say it's outside of the scope of this PR.

controllers/component_controller.go
johnmcollier: I believe that's planned for scm-file-retriever in the future (e.g. detecting private repos and only using token if needed), but not available at the moment.

controllers/component_controller.go
johnmcollier: Haha sure

go.mod
johnmcollier: Dunno, maybe? I'll take a look.

Pulling in scm-file-retriever required us to bump a number of other dependencies (namely cause they use a newer client-go version), so it might be related to that?

controllers/component_controller.go
johnmcollier: Updated.

go.mod
johnmcollier: Updated.

PR #64, opened 2022-03-04
[WIP] Alizer integration with Component Detection

maysunfaisal:

### Work
- integrates the Alizer go module to detect a devfile from the registry if missing

### TODO
- refactor
- tests


config/manager/kustomization.yaml
```diff
@@ -16,5 +16,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/redhat-appstudio/application-service
-  newTag: next
+  newName: maysunfaisal/hasoperator
+  newTag: alizer
```
johnmcollier, line 20: Commenting so that this doesn't get merged in 🙂

pkg/util/util.go
```diff
@@ -149,70 +160,160 @@ func CloneRepo(clonePath, repoURL string) error {
...
 	}
 
-	// if we didnt find any devfile we should return an err
 	if len(devfileMapFromRepo) == 0 && currentLevel == 0 {
-		err = fmt.Errorf("unable to find any devfile(s) in the multi component repo, devfiles can be detected only upto a depth of %v dir", depth)
+		// if we didnt find any devfile we should return an err
+		err = &NoDevfileFound{location: localpath}
+	} else if !isDevfilePresent && currentLevel == depth {
+		// if we didnt find any devfile upto our desired depth, then use alizer
+		devfileBytes, detectedDevfileEndpoint, err := AnalyzeAndDetectDevfile(localpath)
```
johnmcollier, line 227: If we only ever call alizer at the max depth (I’m assuming that’s what localpath is set to), do we risk missing things if we needed to have scanned at a higher depth (if that makes sense)

pkg/util/util.go
```diff
@@ -149,70 +160,160 @@ func CloneRepo(clonePath, repoURL string) error {
...
+		}
+
+	}
+
+	return devfileMapFromRepo, devfilesURLMapFromRepo, err
+}
+
+func getAlizerDevfileTypes() ([]recognizer.DevFileType, error) {
+	types := []recognizer.DevFileType{}
+	registryIndex, err := registryLibrary.GetRegistryIndex(DevfileStageRegistryEndpoint, registryLibrary.RegistryOptions{
```
johnmcollier, line 247: Shouldn’t use stage here

johnmcollier (commented):
No big concerns, mostly comments around formatting.

pkg/util/util.go
maysunfaisal: will configure it as part of a follow up PR

johnmcollier (approved):

maysunfaisal:
Merging now, going to address test enhancement and review enhancements as a separate PR

PR #65, opened 2022-03-08
If 'default' image repo is used, also set up 'default' pull secret

sbose78:
## Problem statement

It is painful to try out the App Studio experience from the UI because the user needs to 'somehow' create the `redhat-appstudio-registry-pull-secret`. This PR removes the need for the same by having the controller *conditionally* create the same upon creation of a Component. 

Note, the controller already sets the the output image to `quay.io/redhat-appstudio/user-workload:unique-tag`, this PR also sets the pull secret for the same. 

## Why

Provide a low-code way of setting up the system-wide 'default' pull secret named `default-user-workload-pull-secret` required for pushing images built out of components when the output image is not specified ( ie, in the UI flows )

### Application Service controller setup

The secret is mounted into the file system of the Application Service controller deployment as  "/etc/build/.dockerconfigjson" . (Needs to be done in the `application-service` namespace). The `/etc/build` Path can be overriden to `/current/working/directory` by setting the `BUILD_CREDENTIALS` env var such that `gitops/testdata` is usable in tests.

### User namespaces

* Upon component creation, conditionally, a pull secret named `component-foo` gets created ( by the controller ) with the contents of the "/etc/build/.dockerconfigjson" .
* This secret is then associated with the build resources for the component.
* This removes the need for the user to set the `redhat-appstudio-registry-pull-secret` out of band.

## Deployment

Before deployment of this change on any environment,
* Create `default-user-workload-pull-secret` in `application-service` associated with `quay.io/redhat-appstudio/user-workload`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,16 +53,6 @@ The following section outlines the steps to deploy HAS on a physical Kubernetes 
 * Install `OpenShift GitOps` from the in-cluster Operator Marketplace.
 * `oc -n openshift-gitops apply -f https://raw.githubusercontent.com/redhat-appstudio/infra-deployments/main/argo-cd-apps/base/build.yaml`
 
-As a user, upon creation of Component, Tekton resources would be created by the controller. 
-
-If you wish to get 'working' PipelineRuns,
-* Create an image pull secret named `redhat-appstudio-registry-pull-secret`. See [Kubernetes docs](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/#registry-secret-existing-credentials) for more information on how to create
-`Secrets` containing registry credentials. 
-* Configure the default image repository to which Pipelines would push images to by defining the environment variable `IMAGE_REPOSITORY` for the operator deployment.
-Defaults to `quay.io/redhat-appstudio/user-workload`.
-
-Pipelines would use the credentials in the image pull secret `redhat-appstudio-registry-pull-secret` to push to $IMAGE_REPOSITORY.
-
 
 
 ### Creating a GitHub Secret for HAS
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -24,8 +24,16 @@ spec:
     spec:
       securityContext:
         runAsNonRoot: true
+      volumes:
+      - name: default-user-workload-pull-secret
+        secret:
+          secretName: default-user-workload-pull-secret
       containers:
-      - command:
+      - volumeMounts:
+        - name: default-user-workload-pull-secret
+          mountPath: "/etc/build"
+          readOnly: true
+        command:
         - /manager
         args:
         - --leader-elect
--- a/controllers/component_controller.go
+++ b/controllers/component_controller.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"path"
 	"reflect"
+	"strings"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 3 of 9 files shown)


gitops/testdata/.dockerconfigjson
```diff
@@ -0,0 +1 @@
+{"auths":{"https://index.docker.io/v1/":{"username":"sbose78","password":"foo","auth":"c2Jvc2U3ODpmb28=","email":""}}}
```
sbose78, line 1: some random json here

gitops/generate_build.go
```diff
@@ -97,7 +118,8 @@ func DetermineBuildExecution(component appstudiov1alpha1.Component, params []tek
 			{
 				Name: "registry-auth",
 				Secret: &corev1.SecretVolumeSource{
-					SecretName: "redhat-appstudio-registry-pull-secret",
+					SecretName: component.Name,
```
mmorhun, line 121: Is it supposed to be different for each component or the same for all?

config/manager/manager.yaml
```diff
@@ -24,8 +24,16 @@ spec:
...
         runAsNonRoot: true
+      volumes:
+      - name: default-user-workload-pull-secret
+        secret:
+          secretName: default-user-workload-pull-secret
       containers:
-      - command:
+      - volumeMounts:
+        - name: default-user-workload-pull-secret
+          mountPath: "/etc/build"
```
mmorhun, line 34:
```suggestion
          mountPath: "/etc/build-credentials"
```

gitops/generate_build.go
```diff
@@ -97,7 +118,8 @@ func DetermineBuildExecution(component appstudiov1alpha1.Component, params []tek
...
+					// TODO: Cleanup "redhat-appstudio-registry-pull-secret" from namespaces,
```
mmorhun, line 122: Do we need to care about this before GA?

controllers/component_controller.go
```diff
@@ -328,6 +329,40 @@ func (r *ComponentReconciler) runBuild(ctx context.Context, component *appstudio
...
+		err := controllerutil.SetOwnerReference(component, &defaultSecret, r.Scheme)
+		if err != nil {
+			log.Error(err, fmt.Sprintf("Unable to set owner reference for %v", defaultSecret))
+			return err
+		}
+		secret := &corev1.Secret{}
+		err = r.Get(ctx, types.NamespacedName{Name: defaultSecret.Name, Namespace: component.Namespace}, secret)
+		if err != nil {
+			if errors.IsNotFound(err) {
+				err = r.Client.Create(ctx, &defaultSecret)
```
mmorhun, line 352: When the component gets deleted, we need to clean the secret too.

controllers/component_controller.go
```diff
@@ -328,6 +329,40 @@ func (r *ComponentReconciler) runBuild(ctx context.Context, component *appstudio
...
 
+	// Not exposing this as config on purpose since this is supposed to be moved out eventually
+	// Also, this is really being used for tests.
+	credentialPath := os.Getenv("BUILD_CREDENTIALS")
+	if credentialPath == "" {
+		// expect this to be unset in production.
+		credentialPath = "/etc/build/.dockerconfigjson"
+	}
+
+	if strings.Contains(component.Spec.Build.ContainerImage, r.ImageRepository) {
```
mmorhun, line 340: Is it worth moving the code into a separate function? (maybe include the block above too)

controllers/component_controller.go
```diff
@@ -328,6 +329,40 @@ func (r *ComponentReconciler) runBuild(ctx context.Context, component *appstudio
 
 	log := r.Log.WithValues("Namespace", component.Namespace, "Application", component.Spec.Application, "Component", component.Name)
 
+	// Not exposing this as config on purpose since this is supposed to be moved out eventually
```
mmorhun, line 332: > this is supposed to be moved out eventually

Are you talking about `credentialPath` only or the secret mount in general?

mmorhun:
Thinking about using gitops repository manually (this is a valid use-case, isn't it?). We have a requirement to have created registry credentials secret in order to be able to push built image. Is it ok to require user to create a secret with specific name or would it be better to pass secret name as a parameter to build pipeline? It would be clearer from user perspective, but makes things a bit more complicated for us.

gitops/generate_build.go
sbose78: We'll be in a different cluster probably by then

gitops/generate_build.go
sbose78: The name will be same as that of the component - but the secret content will be the same if registry repo is the default well-known one.

mmorhun: I see. Let's keep it as is

README.md
```diff
@@ -53,16 +53,6 @@ The following section outlines the steps to deploy HAS on a physical Kubernetes
 * Install `OpenShift GitOps` from the in-cluster Operator Marketplace.
 * `oc -n openshift-gitops apply -f https://raw.githubusercontent.com/redhat-appstudio/infra-deployments/main/argo-cd-apps/base/build.yaml`
 
-As a user, upon creation of Component, Tekton resources would be created by the controller.
```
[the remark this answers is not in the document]
sbose78: I'll add some deployment instructions 👍

config/manager/manager.yaml
sbose78: 🤔 why so?

controllers/component_controller.go
sbose78: Yes, it should get cleaned by `ownerReferences`

controllers/component_controller.go
sbose78: Yes 👍

config/manager/manager.yaml
mmorhun: Well, to me `build` folder is more about storing current build artifacts (maybe a habit from `make`). Adding word `credentials` makes it clear.

johnmcollier:
@sbose78 Is this PR needed, or can it be closed?

johnmcollier:
With https://github.com/redhat-appstudio/application-service/pull/96 I think this is no longer needed, so I'm closing this out...

PR #67, opened 2022-03-09
Detect build pipeline to use for component

mmorhun:

#### What this MR does

This MR adds functionality to choose build pipeline depending on the component. It uses data stored in component's model to make decision. The algorithm of pipeline detection is the following:
1. Check component's model for language and framework. If they are specified, use available language specific build pipeline.
2. If the language is not detected, then look for dockerfile and use it
3. If nothing above works, return `noop` build as it is unknow how to build the component.

#### How to test

a. _Java Quarkus scenario_
1. Create a sample application
2. Create a component that points to [Quarkus example](https://github.com/devfile-samples/devfile-sample-code-with-quarkus): 
```yaml
kind: Component
metadata:
  name: component-sample
  namespace: test
spec:
  source:
    git:
      url: https://github.com/devfile-samples/devfile-sample-code-with-quarkus
  build:
    containerImage: docker.io/username/foobar:dev-code-with-quarkus
```
3. Check that a new PipeplineRun for Java projects is created.

b. _Dockerfile scenario_
1. Create a sample application
2. Create a component that points to a repository ([example](https://github.com/mmorhun-org/devfile-stack-go)) with a simple Go lang server, Dockerfile and Devfile:
```golang
package main
import (
  "fmt"
  "net/http"
)
func main() {
  fmt.Println("Starting server on port 8080")
  http.HandleFunc("/", HelloServer)
  http.ListenAndServe("0.0.0.0:8080", nil)
}
func HelloServer(w http.ResponseWriter, r *http.Request) {
  fmt.Fprintf(w, "Hello, %s!", r.URL.Path[1:])
}
```
```dockerfile
FROM golang
COPY main.go main.go
EXPOSE 8080
ENTRYPOINT go run main.go
```
```yaml
schemaVersion: 2.2.0
metadata:
  name: simple-go
  version: 1.1.0
  displayName: Simple Go lang server
  description: Test project for simple Go lang server
  attributes:
    alpha.dockerimage-port: 8080
components:
  - name: outerloop-build
    image:
      imageName: docker.io/mm4eche/foobar:golang
      dockerfile:
        uri: ./Dockerfile
        buildContext: .
        rootRequired: false
```
3. Check that a new PipeplineRun for Devfile projects is created.


gitops/generate_build.go
```diff
@@ -105,8 +104,51 @@ func DetermineBuildExecution(component appstudiov1alpha1.Component, params []tek
...
+	// Check for Dockerfile
+	filterOptions := devfilecommon.DevfileOptions{
+		ComponentOptions: devfilecommon.ComponentOptions{
+			ComponentType: devfilev1alpha2.ImageComponentType,
+		},
+	}
+	devfileComponents, _ := componentDevfileData.GetComponents(filterOptions)
+	for _, component := range devfileComponents {
+		if component.Image != nil && component.Image.Dockerfile != nil {
+			return "docker-build"
```
sbose78, line 146: Could we also get the path to the Dockerfile and pass that on to the Pipeline ? Without that, existing M2 workflows would break.

https://github.com/sbose78/devfile-sample-code-with-quarkus/blob/main/devfile.yaml#L27

sbose78 (changes requested):

gitops/generate_build.go
```diff
@@ -71,18 +75,32 @@ func GenerateBuild(fs afero.Fs, outputFolder string, component appstudiov1alpha1
 	return nil
 }
 
+// GenerateInitialBuildPipelineRun generates pipeline run for initial build of the component.
+func GenerateInitialBuildPipelineRun(component appstudiov1alpha1.Component) tektonapi.PipelineRun {
```
sbose78, line 79: If you separate this out, this won't be counted into code coverage, you'll need to add a small unit test for this :)

sbose78:
Tried out the PR, works great 🥳 , 


I have a couple of changes to suggest:

1. Could we use the buildContext too? https://github.com/sbose78/devfile-sample-code-with-quarkus/blob/main/devfile.yaml#L28 , and pass that on as a parameter to the PipelineRun / TriggerTemplate ( https://github.com/redhat-appstudio/build-definitions/blob/main/pipelines/docker-build.yaml#L22 ) 

2. Could we flip the order of discovery ? If Dockerfile is present, let's go for that. Else, go for runtime-based pipelines. This is because a Dockerfile-based build is more likely to be accurate than a s2i-based build.

jduimovich:
> Could we flip the order of discovery ? If Dockerfile is present, let's go for that. Else, go for runtime-based pipelines. This is because a Dockerfile-based build is more likely to be accurate than a s2i-based build.
FYI - that ordering could break (and does)
In the java example we use, there is a pom.xml and Dockerfile. Building via Dockerfile fails because the maven build has to occur first, the dockerfile assumes a  target was already present.

gitops/generate_build.go
mmorhun: A test for test... ok

sbose78 (approved):

sbose78:
I will merge this after https://github.com/redhat-appstudio/application-service/pull/68

sbose78:
Raise a PR on infra-deployments, please.

PR #69, opened 2022-03-10
Use the built-in pruning mechanism

sbose78:


sbose78:
/approve

self-approving to fix tekton ci

PR #72, opened 2022-03-10
Add releaseStrategies field to component CRD

davidmogar:
Add a new field to the component CRD. This field will be used by the Release service to trigger releases for a given component.


johnmcollier (commented):
+1

PR #77, opened 2022-03-15
Use proper APIVersion for Route

johnmcollier:
Fixes the warning message seen when deploying GitOps resources generated by HAS


sbose78 (approved):
Thanks!

PR #79, opened 2022-03-16
resolve broken docker-build pipeline definition

sbose78:
part of https://github.com/redhat-appstudio/build-definitions/pull/33


sbose78:
Tested the pipeline

image

sbose78:
Self approving and merging this to avoid having a broken staging.

PR #83, opened 2022-03-29
Use build-templates-bundle for pull-request check

Michkov:


Michkov:
@johnmcollier Could you merge this?

PR #86, opened 2022-04-06
Clean up old CDQ resources when they are requeued

johnmcollier:
- Cleans up old CDQ resources when they get deleted
- Also adds new `Detecting` status condition to the CDQs, as discussed earlier
   - Name to be decided, so it's just a placeholder til we finalize on a proper name for the status.
- Test case for the CDQ deletion logic.


elsony:
@rohitkrai03 We are adding a new "Processing" status in the CDQ to avoid some cases that cause OOM problems in HAS. This is a heads up just in case HAC needs to react to it.

maysunfaisal (approved):

johnmcollier:
This bumps up code coverage by half a percentage and codecov is still complaining 😆

PR #87, opened 2022-04-07
Separate gitops and build stuff from Application Service

mmorhun:
This PR moves gitops generation and build resources creation into [separate repository](https://github.com/redhat-appstudio/build-service).


mmorhun:
The PR is ready for reviews, but please do not merge until other PRs are ready. I'll notify you then.

gitops/gitops.go
```diff
@@ -21,6 +21,7 @@ import (
 	"path/filepath"
```
johnmcollier, line 21: Does this file (gitops.go) not need to be extracted out too?

gitops/gitops.go
```diff
@@ -58,7 +59,7 @@ func GenerateAndPush(outputPath string, remote string, component appstudiov1alph
 
 	// Generate the gitops resources
 	componentPath := filepath.Join(repoPath, context, "components", componentName, "base")
-	if err := Generate(appFs, componentPath, component); err != nil {
+	if err := gitopsgen.Generate(appFs, componentPath, component); err != nil {
```
maysunfaisal, line 62: one other suggestion that i have is, if we are moving gitops to it own repo; how about `build-service` `Generate()` func have its own parameters instead of taking in a `component` type from HAS? Maybe `build-service``Generate()` can have a struct that takes in the necessary values and be independent rather than relying on HAS `component` type 🤔 

What do you think?

controllers/suite_test.go
```diff
@@ -77,8 +77,8 @@ var _ = BeforeSuite(func() {
 		CRDDirectoryPaths: []string{
 			filepath.Join("..", "config", "crd", "bases"),
 			filepath.Join("..", "hack", "routecrd"),
-			filepath.Join(build.Default.GOPATH, "pkg", "mod", "github.com", "tektoncd", "triggers@v0.17.1", "config"),
-			filepath.Join(build.Default.GOPATH, "pkg", "mod", "github.com", "tektoncd", "pipeline@v0.32.1", "config")},
+			filepath.Join(build.Default.GOPATH, "pkg", "mod", "github.com", "tektoncd", "triggers@v0.19.1", "config"),
```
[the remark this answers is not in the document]
mmorhun: My bad, forgot to remove. Thanks.

gitops/gitops.go
mmorhun: 1. `build-service` operator depends on `component` anyway, because it watches them in the cluster.
2. We use different fields of `component` in gitops resources generation, so the type will have to mimic many `component` fields (except auxiliary k8s fields). And that should be done manually. Another downside is that we might change the logic of gitops generation and require new fields or abandon old. But if we move gitops out from build service in a separate repository, then the circular dependency goes away.

I am not sure which option is better. @sbose78 what do you think?

gitops/gitops.go
johnmcollier: So the interactions with git will remain within HAS?

controllers/suite_test.go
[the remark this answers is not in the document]
johnmcollier: 👍

gitops/gitops.go
mmorhun: Yes, it will, as the [issue](https://issues.redhat.com/browse/PLNSRVCE-138) say.

johnmcollier:
@mmorhun The CI is throwing the following error for go mod tidy:
```
Run go mod tidy
github.com/redhat-appstudio/application-service imports
	github.com/tektoncd/triggers/pkg/apis/triggers/v1alpha1 imports
	github.com/google/cel-go/cel imports
	github.com/google/cel-go/parser imports
	github.com/antlr/antlr4/runtime/Go/antlr loaded from github.com/antlr/antlr4/runtime/Go/antlr@v0.0.0-20210826220005-b48c857c3a0e,
	but go 1.16 would fail to locate it:
	ambiguous import: found package github.com/antlr/antlr4/runtime/Go/antlr in multiple modules:
	github.com/antlr/antlr4 v0.0.0-20200503195918-621b933c7a7f (/home/runner/go/pkg/mod/github.com/antlr/antlr4@v0.0.0-20200503195918-621b933c7a7f/runtime/Go/antlr)
	github.com/antlr/antlr4/runtime/Go/antlr v0.0.0-20210826220005-b48c857c3a0e (/home/runner/go/pkg/mod/github.com/antlr/antlr4/runtime/!go/antlr@v0.0.0-20210826220005-b48c857c3a0e)

To proceed despite packages unresolved in go 1.16:
	go mod tidy -e
If reproducibility with go 1.16 is not needed:
	go mod tidy -compat=1.17
For other options, see:
	https://golang.org/doc/modules/pruning
Error: Process completed with exit code 1.
```

johnmcollier:
@mmorhun Just following up once more on the error message seen in the CI ☝️

mmorhun:
Thank you @johnmcollier . Currently I'm working on alternate solution and if it is successful, then I'll close this PR and open a new one instead.

mmorhun:
Closing on favor of https://github.com/redhat-appstudio/application-service/pull/96

PR #92, opened 2022-04-19
[DEVHAS-26] feat: Support container image components

johnmcollier:
Fixes https://issues.redhat.com/browse/DEVHAS-26

This PR implements support for (public and private) container image components in HAS. It also moves the `spec.Source.GitSecret` field to `spec.Secret` to allow both Git and Image components to reference secrets.

When a Component is created in HAS referencing a container image, like so:
```
kind: Component
metadata:
  name: component-example
  namespace: default
spec:
  application: application-example
  build:
    containerImage: ''
  resources: {}
  source:
    image:
      containerImage: docker.io/openshift/hello-openshift
```

HAS will generate gitops resources that reference that container image and push it to the gitops repository. If a private container image needs to be used, then the `spec.secret` field must be set to a valid Kubernetes `ImagePullSecret` containing credentials for the image registry containing the private container image.

**Note:** Since no build will occur, the Tekton pipeline resources will not be generated, nor will any build resources be created cluster-side.

### DIFF
--- a/pkg/devfile/devfile.go
+++ b/pkg/devfile/devfile.go
@@ -16,6 +16,7 @@
 package devfile
 
 import (
+	"github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
 	"github.com/devfile/api/v2/pkg/attributes"
 	"github.com/devfile/api/v2/pkg/devfile"
 	devfilePkg "github.com/devfile/library/pkg/devfile"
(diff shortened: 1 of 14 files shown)


pkg/devfile/devfile.go
```diff
@@ -90,6 +91,37 @@ func ConvertApplicationToDevfile(hasApp appstudiov1alpha1.Application, gitOpsRep
 	return devfileData, nil
 }
 
+func ConvertImageComponentToDevfile(comp appstudiov1alpha1.Component) (data.DevfileData, error) {
```
maysunfaisal, line 94: could you pls provide a small func desc

pkg/devfile/devfile_test.go
```diff
@@ -146,6 +146,69 @@ func TestConvertApplicationToDevfile(t *testing.T) {
 	}
 }
 
+func TestConvertImageComponentToDevfile(t *testing.T) {
+	//devfileAttributes := attributes.Attributes{}.PutString("gitOpsRepository.url", "https://github.com/testorg/petclinic-gitops").PutString("appModelRepository.url", "https://github.com/testorg/petclinic-app"),
```
maysunfaisal, line 150: comment

maysunfaisal (approved):
minor nits only, leave it up to you if you want to change them nw or  later

maysunfaisal (approved):

johnmcollier:
Will merge now to unblock you @maysunfaisal, will address the review comments in a follow up pr.

PR #93, opened 2022-04-20
update has to use PVC instead of volumeClaimTemplate

mbharatk:
### What does this PR do?
As PVC is created for the whole namespace and the pipelineRun is using subPath, the PVC was being full very soon, and to overcome this, as a workaround we used volumeClaimTemplate, which creates PVC for each pipelineRun.

Now, there is a [PVC-cleaner](https://github.com/redhat-appstudio/pvc-cleaner) which will delete the subPath when a pipelineRun is deleted.
This PR is to switch back to subPaths from volumeClaimTemplate.

### What issues does this PR fix or reference?
https://issues.redhat.com/browse/PLNSRVCE-174

### DIFF
--- a/.tekton/pull-request.yaml
+++ b/.tekton/pull-request.yaml
@@ -20,10 +20,6 @@ spec:
   serviceAccountName: pipeline
   workspaces:
     - name: workspace
-      volumeClaimTemplate:
-        spec:
-          accessModes:
-            - ReadWriteOnce
-          resources:
-            requests:
-              storage: 200Mi
+      persistentVolumeClaim:
+        claimName: app-studio-default-workspace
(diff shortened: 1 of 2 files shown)


johnmcollier:
@mbharatk With the build service component being removed from HAS and into its own repo, should this go in https://github.com/redhat-appstudio/build-service instead? (Or as well?)

mbharatk:
@johnmcollier yes, it is already added to the build-service repo as well 
Link to build-service PR: https://github.com/redhat-appstudio/build-service/pull/5

johnmcollier:
@mbharatk Ah, thanks. I missed that.

johnmcollier (approved):

PR #94, opened 2022-04-20
[DEVHAS-94] Requeue the Component when its Application can't be found

johnmcollier:
Fixes https://issues.redhat.com/browse/DEVHAS-94

Updates the Component controller logic to requeue if a Component's Application cannot be found, or is not finished reconciling.


maysunfaisal (approved):

maysunfaisal:
@johnmcollier do we want a simple test where we create a component before an application?

johnmcollier:
@maysunfaisal I'll update the existing test case (#4) that covered this scenario

johnmcollier:
@maysunfaisal Updated, thanks!

maysunfaisal (approved):

PR #96, opened 2022-04-22
Delete creation of build objects

mmorhun:
This PR deletes direct creation of build objects from `Component` controller.
Also updates some deps to newer version.


johnmcollier:
@mmorhun Can you also remove the tekton CRDs from https://github.com/redhat-appstudio/application-service/blob/main/controllers/suite_test.go#L80-L81

johnmcollier (commented):
Generally looks good to me. I think there's a couple more sections that can be removed as well:

https://github.com/redhat-appstudio/application-service/blob/a118e0993483f79ca4a113bb43ef5aadf7ff45d7/controllers/component_controller.go#L387-L406

https://github.com/redhat-appstudio/application-service/blob/a118e0993483f79ca4a113bb43ef5aadf7ff45d7/main.go#L92-L105

https://github.com/redhat-appstudio/application-service/blob/a118e0993483f79ca4a113bb43ef5aadf7ff45d7/main.go#L119-L122

https://github.com/redhat-appstudio/application-service#setting-up-the-appstudio-build-service-environment

sbose78:
/hold 

Since we are close to the demo

mmorhun:
@johnmcollier thank you for the review.
I'll get rid of `triggersapi` and `taskrunapi`, but I'm afraid we should keep image repository and the webhook route as they are parts of the component: 
- image repository is a field in component CR and we should make it mandatory of provide a default value in order to have the component configuration complete (gitops generation will fail otherwise).
- build webhook route is displayed in component status (and naturally belongs to it). As Application Service operator owns the `Component` resource, only it should change the status. So, we cannot move the route fetching anywhere else.

johnmcollier:
@mmorhun responses inline

> image repository is a field in component CR and we should make it mandatory of provide a default value in order to have the component configuration complete (gitops generation will fail otherwise).

Okay, that makes sense. Let's keep `IMAGE_REPOSITORY`. 

> build webhook route is displayed in component status (and naturally belongs to it). As Application Service operator owns the Component resource, only it should change the status. So, we cannot move the route fetching anywhere else.

So I understand this, but I have two further questions:
1) We only get the EL Route after build resource creation has succeeded (see https://github.com/redhat-appstudio/application-service/blob/main/controllers/component_controller.go#L384), but if HAS no longer initiates the build resource creation, how will it know when it's succeeded?

2) Why does the EL Route need to be part of the Component status?

mmorhun:
@johnmcollier 
> We only get the EL Route after build resource creation has succeeded (see https://github.com/redhat-appstudio/application-service/blob/main/controllers/component_controller.go#L384), but if HAS no longer initiates the build resource creation, how will it know when it's succeeded?

Via a watcher (or by retry reconcile as it is done now).
In fact, build resource creation are initiated by HAS, but done by Argo CD. To my mind, it's HAS responsibility to control what it has initiated and delegated.

> Why does the EL Route need to be part of the Component status?

In my opinion, the build webhook is a part of `Component` current state. As a user, I'd look into my component resource get info about it in the spec and status. As the `Component` is owned by HAS, only it can manage its status.

johnmcollier:
Okay, that sounds fine.

PR #106, opened 2022-04-30
Allow the overriding of the default Tekton bundle

brunoapimentel:
[PLNSRVCE-141](https://issues.redhat.com//browse/PLNSRVCE-141)

This is a reimplementation of #74 considering the splitting of the code into the build-service operator.

----

This PR provides a simple mechanism to allow the overriding of the Tekton bundle that will be inserted into the generated TriggerTemplate when a Component CR is created.

To do that, a configmap needs to be created in the same namespace as the Component CR:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: build-pipelines-defaults
data:
  default_build_bundle: "quay.io/user/custombundle:version" 
```

In case this configmap is not available, HAS will try to search for one with the same name at the `build-templates` namespace. If no configmap is found, it will resort to the one that's currently [hardcoded](https://github.com/brunoapimentel/application-service/blob/96308c90f86acc7b6993a4deb7cf5da17835a3e1/gitops/generate_build.go#L42).

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -84,3 +84,11 @@ rules:
   - get
   - patch
   - update
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  verbs:
+  - get
+  - list
+  - watch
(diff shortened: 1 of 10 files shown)


brunoapimentel:
This PR was tested in a real cluster using the following steps:
- Create an Application CR
- Create a Component CR
- Check the Tekton bundle that was pushed to the generated gitops repository and see if it matches the one specified

The following scenarios were tested:
- No configmap (fallback bundle)
- Configmap at the "build-definitions" namespace 
- Configmap in the same namespace as the Component

mmorhun (approved):
LGTM

PR #110, opened 2022-05-05
WIP: PLNSRVCE-201: add shared resource volume for push secret to build pipeline run

gabemontero:
to go with the build-definition work @Michkov is doing for consuming the shared resource shared secret for push secret

part of https://issues.redhat.com/browse/PLNSRVCE-201

@sbose78 FYI the "replacement" for https://github.com/redhat-appstudio/application-service/pull/65

### DIFF
--- a/gitops/generate_build.go
+++ b/gitops/generate_build.go
@@ -97,11 +97,29 @@ func getInitialBuildWorkspaceSubpath() string {
 // in webhooks-triggered pipelineRuns as well as user-triggered PipelineRuns
 func DetermineBuildExecution(component appstudiov1alpha1.Component, params []tektonapi.Param, workspaceSubPath string) tektonapi.PipelineRunSpec {
 
+	truVar := true
 	pipelineRunSpec := tektonapi.PipelineRunSpec{
 		Params: params,
 		PipelineRef: &tektonapi.PipelineRef{
 			Name:   determineBuildPipeline(component),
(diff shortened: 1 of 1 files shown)


gitops/generate_build.go
```diff
@@ -97,13 +97,31 @@ func getInitialBuildWorkspaceSubpath() string {
...
+		PodTemplate: &tektonapi.PodTemplate{
+			Volumes: []corev1.Volume{
+				{
+					Name: "registry-auth-volume",
+					VolumeSource: corev1.VolumeSource{
+						CSI: &corev1.CSIVolumeSource{
+							ReadOnly: &truVar,
+							Driver:   "csi.sharedresource.openshift.io",
+							VolumeAttributes: map[string]string{
+								"sharedSecret": "appstudio-push-secret",
```
Michkov, line 117: Nope this secret is not for been used by this, maybe we can rename it. Currently we have sharedSecret for Appstudio services to be used by Pipelines-as-Code ci, not by application-service. For application service we need to create another shared secret which will be allowed to be used by whole cluster.

gabemontero: hmm ... I see 

so minimally I still need to increase my understanding of the end to end picture 

and most likely am not getting all the nuance behind @sbose78 's https://github.com/redhat-appstudio/application-service/pull/65

PR #117, opened 2022-05-17
Update component spec with default image if the image is not set

mmorhun:

This MR explicitly sets default component image in component spec if it was not set manually.
Such behavior:
- Makes it more clear for the user which image is used and how to change it.
- Analysis of the component (including gitops creation) is made from actual component configuration in the cluster, not a modified in-memory copy.


Michkov:
Fix of [PLNSRVCE-291](https://issues.redhat.com/browse/PLNSRVCE-291)

Michkov (approved):

johnmcollier (approved):

PR #128, opened 2022-05-31
Update HAS to use Alizer DetectComponent instead of Analyze

yangcao77:
This PR updates HAS to use Alizer `DetectComponent` instead of `Analyze`.

`DetectComponent` fix the issue with detecting dotnet language, since it use the config file to order the detected languages, while `Analyze` order the result using language percentage

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -169,6 +169,12 @@ deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in
 undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/config.
 	$(KUSTOMIZE) build config/default | kubectl delete -f -
 
+deploy-kcp: install ## Install CRDs and deploy HAS on KCP
+	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
+	GITHUB_ORG=${GITHUB_ORG} DEVFILE_REGISTRY_URL=${DEVFILE_REGISTRY_URL} $(KUSTOMIZE) build config/kcp | kubectl apply -f -
+
+undeploy-kcp: # Undeploy HAS from KCP (including CRDs)
+	$(KUSTOMIZE) build config/kcp | kubectl delete -f -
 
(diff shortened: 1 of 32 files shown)


pkg/devfile/detect.go
```diff
@@ -233,24 +233,33 @@ func (a AlizerClient) Analyze(path string) ([]language.Language, error) {
...
 	if err != nil {
 		return nil, "", "", err
 	}
 
-	for _, language := range alizerLanguages {
+	if len(alizerComponents) == 0 {
+		return nil, "", "", &NoDevfileFound{Location: path}
+	}
+
+	// Assuming it's a single component. as multi-component should be handled before
```
maysunfaisal, line 261: so what does this mean? `DetectComponents` can return multple components and each component have different languages?

maysunfaisal: Because previously AlizerLanguage was an array []Language{} but now AlizerComponent is an array where Language is inside - []Component{{path, []Language}}

so what are the possible values for AlizerComponent, can you give an example? Since we are taking the index 0 always

yangcao77: for example, running `DetectComponent` against a single component project, the output will be:
```
[
 {
   path: /Users/stephanie/Documents/devfile-samples/devfile-sample-dotnet-basic/ 
   languages: [{C# [csharp dotnet] 15 [net6.0] [] true} {JavaScript [js node nodejs] 75 [] [] true} {Smalltalk [squeak] 15 [] [] false} {GCC Machine Description [] 10 [] [] false}] 
  }
 ]
```

yangcao77: The output for languages have no differences here.

yangcao77: we have discussed in the meeting. currently we have multi-component field in the spec, and after [the issue ](https://issues.redhat.com/projects/DEVHAS/issues/DEVHAS-80) is in, We will detect if the project is a multi component project when we first scanning through the project. 
so when come to the analyze & look for devfile stage, we always assume it's a single component .

PR #129, opened 2022-06-01
PLNSRVCE-251: fix matching of protected tags

Michkov:
Default repostory should be protected against pushing from other
namespaces. The image tag name should start with namespace.

### DIFF
--- a/gitops/generate_build.go
+++ b/gitops/generate_build.go
@@ -93,9 +93,12 @@ func GenerateBuild(fs afero.Fs, outputFolder string, component appstudiov1alpha1
 }
 
 // GenerateInitialBuildPipelineRun generates pipeline run for initial build of the component.
-func GenerateInitialBuildPipelineRun(component appstudiov1alpha1.Component, gitopsConfig prepare.GitopsConfig) tektonapi.PipelineRun {
+func GenerateInitialBuildPipelineRun(component appstudiov1alpha1.Component, gitopsConfig prepare.GitopsConfig) (tektonapi.PipelineRun, error) {
 	// normalizeOutputImageURL is not called with initial builds so we can ignore the error here
-	params, _ := getParamsForComponentBuild(component, true)
+	params, err := getParamsForComponentBuild(component, true)
+	if err != nil {
+		return tektonapi.PipelineRun{}, err
+	}
 	initialBuildSpec := DetermineBuildExecution(component, params, getInitialBuildWorkspaceSubpath(), gitopsConfig)
 
 	return tektonapi.PipelineRun{
@@ -105,7 +108,7 @@ func GenerateInitialBuildPipelineRun(component appstudiov1alpha1.Component, gito
 			Labels:       getBuildCommonLabelsForComponent(&component),
 		},
 		Spec: initialBuildSpec,
-	}
+	}, nil
 }
 
 func getInitialBuildWorkspaceSubpath() string {
--- a/gitops/generate_build.go
+++ b/gitops/generate_build.go
@@ -191,7 +191,18 @@ func determineBuildPipeline(component appstudiov1alpha1.Component) string {
 	return "noop"
 }
 
-func normalizeOutputImageURL(outputImage, namespace string) (string, error) {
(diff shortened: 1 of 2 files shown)


gitops/generate_build.go
```diff
@@ -232,10 +233,13 @@ func getParamsForComponentBuild(component appstudiov1alpha1.Component, isInitial
...
+
+	if err = protectDefaultImageRepo(outputImage, component.Namespace); err != nil {
+		return []tektonapi.Param{}, err
+	}
+
 	if !isInitialBuild {
-		if outputImage, err = normalizeOutputImageURL(component.Spec.ContainerImage, component.Namespace); err != nil {
-			return []tektonapi.Param{}, err
-		}
+		outputImage = normalizeOutputImageURL(component.Spec.ContainerImage)
```
gabemontero, line 242: for my education @Michkov - why do we not want to surface the error capturing the invalid user tag combination?

don't we want to invalidate the entire component interaction here?

gabemontero (commented):
I have one question / *possible* concern @Michkov 

thanks

gabemontero (approved):

gitops/generate_build.go
```diff
@@ -232,10 +233,13 @@ func getParamsForComponentBuild(component appstudiov1alpha1.Component, isInitial
 	sourceCode := component.Spec.Source.GitSource.URL
 	outputImage := component.Spec.ContainerImage
 	var err error
+
+	if err = protectDefaultImageRepo(outputImage, component.Namespace); err != nil {
+		return []tektonapi.Param{}, err
```
psturc, line 238: I just noticed that we don't check the error value returned from this function in [GenerateInitialBuildPipelineRun function](https://github.com/redhat-appstudio/application-service/blob/632e5372c62f42ac0e31625fec32f7022c36e520/gitops/generate_build.go#L98), which will result in ignoring the error (for example in case the `.spec.containerImage` is set to `quay.io/redhat-appstudio/user-workload`) and initiating the build anyway.

@Michkov do you think we should update that?

gabemontero: yeah either we update https://github.com/redhat-appstudio/application-service/blob/632e5372c62f42ac0e31625fec32f7022c36e520/gitops/generate_build.go#L98 to no longer ignore the error, or we move the call to `protectDefaultImageRepo` here after the is initial build check immediately below

unless there is another consideration @psturc and I were missing when he and I discussed his finding

Michkov: updated, now returning error from GenerateInitialBuildPipelineRun

psturc: LGTM 👍

psturc:
/lgtm

so after the merge, apart from updating the application-service reference in build-service, I guess we'll also have to update the [GenerateInitialBuildPipelineRun function call](https://github.com/redhat-appstudio/build-service/blob/a093143fef03c50f80203d86f773102a4582432d/controllers/component_build_controller.go#L219) and perhaps add a new test [here](https://github.com/redhat-appstudio/build-service/blob/e7bb6676170e5c132b1840b83c0a6d66ac07348b/controllers/component_build_controller_test.go#L141).

Does that sound like reasonable thing to do @Michkov, @gabemontero?

Michkov:
@psturc yes

gabemontero:
> /lgtm
> 
> so after the merge, apart from updating the application-service reference in build-service, I guess we'll also have to update the [GenerateInitialBuildPipelineRun function call](https://github.com/redhat-appstudio/build-service/blob/a093143fef03c50f80203d86f773102a4582432d/controllers/component_build_controller.go#L219) and perhaps add a new test [here](https://github.com/redhat-appstudio/build-service/blob/e7bb6676170e5c132b1840b83c0a6d66ac07348b/controllers/component_build_controller_test.go#L141).
> 
> Does that sound like reasonable thing to do @Michkov, @gabemontero?

yes / +1

gabemontero:
let's merge @Michkov @psturc , yes?

gabemontero:
ok @johnmcollier we are good to go now

johnmcollier (approved):

PR #134, opened 2022-06-20
[DEVHAS-101] Application Service Binding Controller

maysunfaisal:
This PR implements:
- controller for ApplicationSnapshotEnvironmentBinding
- ApplicationSnapshotEnvironmentBinding tests
- Updates Readme

#### To test it out:
- Create app `k create -f application/app.yaml`
- Create 2 components `component-sample` & `component-sample2` as per the binding spec `k create -f component/component-basic.yaml`
- Create snapshot `k apply -f snapshot/snapshot.yaml`
- Create binding `k apply -f binding/binding.yaml`

After the binding is created, describe the binding and check the status to see if the GitOps repo refresh was successful and check the GitOps repo for the overlays dir on both the components

Happy reviewing 😄

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -143,7 +143,7 @@ build: generate fmt vet ## Build manager binary.
 run: manifests generate fmt vet ## Run a controller from your host.
 	go run ./main.go
 
-docker-build: test ## Build docker image with the manager.
+docker-build: #test ## Build docker image with the manager.
 	docker build --build-arg ENABLE_WEBHOOKS=${ENABLE_WEBHOOKS} -t ${IMG} .
 
 docker-push: ## Push docker image with the manager.
--- a/Makefile
+++ b/Makefile
@@ -143,7 +143,7 @@ build: generate fmt vet ## Build manager binary.
 run: manifests generate fmt vet ## Run a controller from your host.
 	go run ./main.go
 
-docker-build: #test ## Build docker image with the manager.
+docker-build: test ## Build docker image with the manager.
 	docker build --build-arg ENABLE_WEBHOOKS=${ENABLE_WEBHOOKS} -t ${IMG} .
 
 docker-push: ## Push docker image with the manager.
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,8 @@ GITHUB_ORG ?= redhat-appstudio-appdata
 DEVFILE_REGISTRY_URL ?= https://registry.devfile.io
 ENABLE_WEBHOOKS ?= true
 
+GITOPS_SHARED_CRD = https://raw.githubusercontent.com/redhat-appstudio/managed-gitops/main/appstudio-shared/manifests/appstudio-shared-customresourcedefinitions.yaml
+
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
 CRD_OPTIONS ?= "crd:trivialVersions=true,preserveUnknownFields=false"
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
@@ -162,6 +164,7 @@ install-kcp: manifests kustomize
 
 install: manifests kustomize #install-cert ## Install CRDs into the K8s cluster specified in ~/.kube/config.
 	$(KUSTOMIZE) build config/crd | kubectl apply -f -
+	kubectl apply -f $(GITOPS_SHARED_CRD)
 
 uninstall: manifests kustomize #uninstall-cert ## Uninstall CRDs from the K8s cluster specified in ~/.kube/config.
 	$(KUSTOMIZE) build config/crd | kubectl delete -f -
@@ -276,8 +279,7 @@ apply-crds:
 	-f $(APPLICATIONS_CRD) \
 	-f $(COMPONENT_DETECTION_QUERIES_CRD) \
 	-f $(COMPONENT_CRD) \
-	-f https://raw.githubusercontent.com/redhat-appstudio/managed-gitops/main/appstudio-shared/manifests/appstudio-shared-customresourcedefinitions.yaml
-
+	-f $(GITOPS_SHARED_CRD)
 .PHONY: debug
 debug: dlv generate manifests kustomize apply-crds
 	$(MAKE) debug-stop; \
--- a/Makefile
+++ b/Makefile
@@ -275,7 +275,8 @@ apply-crds:
 	kubectl apply \
 	-f $(APPLICATIONS_CRD) \
 	-f $(COMPONENT_DETECTION_QUERIES_CRD) \
-	-f $(COMPONENT_CRD)
+	-f $(COMPONENT_CRD) \
+	-f https://raw.githubusercontent.com/redhat-appstudio/managed-gitops/main/appstudio-shared/manifests/appstudio-shared-customresourcedefinitions.yaml
 
 .PHONY: debug
 debug: dlv generate manifests kustomize apply-crds
--- a/config/samples/snapshot/snapshot.yaml
+++ b/config/samples/snapshot/snapshot.yaml
@@ -1,0 +1,14 @@
+# API type definition is in github.com/redhat-appstudio/managed-gitops
+apiVersion: appstudio.redhat.com/v1alpha1
+kind: ApplicationSnapshot
+metadata:
+  name: applicationsnapshot-sample
+spec:
+  application: application-sample
+  displayName: my-snapshot
+  displayDescription: my first snapshot
+  components:
+  - name: component-sample
+    containerImage: quay.io/redhat-appstudio/user-workload:application-service-system-component-sample
+  - name: component-sample2
+    containerImage: quay.io/redhat-appstudio/user-workload:application-service-system-component-sample2
--- a/controllers/applicationsnapshotenvironmentbinding_controller.go
+++ b/controllers/applicationsnapshotenvironmentbinding_controller.go
@@ -166,10 +166,11 @@ func (r *ApplicationSnapshotEnvironmentBindingReconciler) Reconcile(ctx context.
 
 		err = gitops.GenerateOverlaysAndPush(tempDir, clone, gitOpsRemoteURL, component, applicationName, environmentName, imageName, appSnapshotEnvBinding.Namespace, r.Executor, r.AppFS, gitOpsBranch, gitOpsContext, componentGeneratedResources)
 		if err != nil {
-			log.Error(err, fmt.Sprintf("unable to get generate gitops resources for %s %v", componentName, req.NamespacedName))
+			gitOpsErr := util.SanitizeErrorMessage(err)
+			log.Error(gitOpsErr, fmt.Sprintf("unable to get generate gitops resources for %s %v", componentName, req.NamespacedName))
 			r.AppFS.RemoveAll(tempDir) // not worried with an err, its a best case attempt to delete the temp clone dir
-			r.SetConditionAndUpdateCR(ctx, &appSnapshotEnvBinding, err)
-			return ctrl.Result{}, err
+			r.SetConditionAndUpdateCR(ctx, &appSnapshotEnvBinding, gitOpsErr)
+			return ctrl.Result{}, gitOpsErr
 		}
 
 		if !isStatusUpdated {
--- a/controllers/applicationsnapshotenvironmentbinding_controller.go
+++ b/controllers/applicationsnapshotenvironmentbinding_controller.go
@@ -1,0 +1,214 @@
+/*
+Copyright 2021-2022 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+	"path/filepath"
+
+	"github.com/go-logr/logr"
+	appstudiov1alpha1 "github.com/redhat-appstudio/application-service/api/v1alpha1"
+	"github.com/redhat-appstudio/application-service/gitops"
+	"github.com/redhat-appstudio/application-service/pkg/util/ioutils"
+	appstudioshared "github.com/redhat-appstudio/managed-gitops/appstudio-shared/apis/appstudio.redhat.com/v1alpha1"
+	"github.com/spf13/afero"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// ApplicationSnapshotEnvironmentBindingReconciler reconciles a ApplicationSnapshotEnvironmentBinding object
+type ApplicationSnapshotEnvironmentBindingReconciler struct {
+	client.Client
+	Scheme   *runtime.Scheme
+	Log      logr.Logger
+	AppFS    afero.Afero
+	Executor gitops.Executor
+	GitToken string
+}
+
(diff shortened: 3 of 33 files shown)


controllers/applicationsnapshotenvironmentbinding_controller.go
```diff
@@ -0,0 +1,241 @@
...
+		return ctrl.Result{}, fmt.Errorf("unable to create temp directory for gitops resources due to error: %v", err)
+	}
+
+	// Get the Snapshot CR
+	appSnapshot := appstudioshared.ApplicationSnapshot{}
+	err = r.Get(ctx, types.NamespacedName{Name: snapshotName, Namespace: appSnapshotEnvBinding.Namespace}, &appSnapshot)
+	if err != nil {
+		log.Error(err, fmt.Sprintf("unable to get the Application Snapshot %s %v", snapshotName, req.NamespacedName))
+		r.SetCreateConditionAndUpdateCR(ctx, &appSnapshotEnvBinding, err)
+		return ctrl.Result{}, nil
```
johnmcollier, line 101: We should requeue here (along with any other instances where we're not requeueing in the controller)

Returning err instead of nil will trigger the requeue.

controllers/applicationsnapshotenvironmentbinding_controller.go
```diff
@@ -0,0 +1,241 @@
...
+			err := fmt.Errorf("application snapshot %s did not reference component %s", snapshotName, componentName)
+			log.Error(err, "")
+			r.SetCreateConditionAndUpdateCR(ctx, &appSnapshotEnvBinding, err)
+			return ctrl.Result{}, nil
+		}
+
+		gitopsStatus := hasComponent.Status.GitOps
+		// Get the information about the gitops repository from the Component resource
+		gitOpsURL = gitopsStatus.RepositoryURL
+		if gitOpsURL == "" {
```
yangcao77, line 153: can we add unit tests on  empty `gitOpsURL`, empty `context`, and non-empty `branch`? for test coverage

config/samples/snapshot/snapshot.yaml
```diff
@@ -0,0 +1,14 @@
+# API type definition is in github.com/redhat-appstudio/managed-gitops
+apiVersion: appstudio.redhat.com/v1alpha1
+kind: ApplicationSnapshot
+metadata:
+  name: applicationsnapshot-sample
+spec:
+  application: application-sample
+  displayName: my-snapshot
+  displayDescription: my first snapshot
+  components:
```
johnmcollier, line 10: What if the same component is referenced more than once here? Is that valid / Do we check for that and validate?

controllers/applicationsnapshotenvironmentbinding_controller.go
```diff
@@ -0,0 +1,241 @@
...
+			r.SetCreateConditionAndUpdateCR(ctx, &appSnapshotEnvBinding, err)
+			return ctrl.Result{}, nil
+		}
+
+		var clone bool
+		var imageName string
+
+		for _, snapshotComponent := range appSnapshot.Spec.Components {
+			if snapshotComponent.Name == componentName {
+				imageName = snapshotComponent.ContainerImage
```
johnmcollier, line 139: Should we `break` here rather than continuing to loop?

controllers/applicationsnapshotenvironmentbinding_controller.go
```diff
@@ -0,0 +1,241 @@
...
+		gitopsStatus := hasComponent.Status.GitOps
+		// Get the information about the gitops repository from the Component resource
+		gitOpsURL = gitopsStatus.RepositoryURL
+		if gitOpsURL == "" {
+			err := fmt.Errorf("unable to create gitops resource, GitOps Repository not set on component status")
+			log.Error(err, "")
+			r.SetCreateConditionAndUpdateCR(ctx, &appSnapshotEnvBinding, err)
+			return ctrl.Result{}, nil
+		}
+		if gitopsStatus.Branch != "" {
```
johnmcollier, line 159: Should we have some common utils function to handle these default values for Branch/Context? IIRC, we have similar logic in either the Application/Component controller.

controllers/applicationsnapshotenvironmentbinding_controller.go
```diff
@@ -0,0 +1,241 @@
...
+		} else {
+			gitOpsBranch = "main"
+		}
+		if gitopsStatus.Context != "" {
+			gitOpsContext = gitopsStatus.Context
+		} else {
+			gitOpsContext = "/"
+		}
+
+		if remoteURL == "" {
```
yangcao77, line 170: I'm wondering what will happen if the `remoteURL` is being set by previous component in the loop?

controllers/applicationsnapshotenvironmentbinding_controller.go
```diff
@@ -0,0 +1,241 @@
...
+			// Construct the remote URL for the gitops repository
+			parsedURL, err := url.Parse(gitOpsURL)
```
johnmcollier, line 172: Similar as above, same logic that's in the Component controller. Should we extract out to a utils function?

gitops/generate.go
```diff
@@ -85,6 +87,31 @@ func Generate(fs afero.Afero, gitOpsFolder string, outputFolder string, componen
 	return GenerateParentKustomize(fs, gitOpsFolder, commonStorage)
 }
 
+// GenerateOverlays generates the overlays dir during the binding reconcile
```
johnmcollier, line 90: Nit: Don't refer to the binding reconcile here, maybe just something like "GenerateOverlays generates the overlays dir from a given BindingComponent"?

gitops/gitops.go
```diff
@@ -76,7 +77,52 @@ func GenerateAndPush(outputPath string, remote string, component appstudiov1alph
...
+	if out, err := e.Execute(repoPath, "git", "add", "."); err != nil {
+		return fmt.Errorf("failed to add files for component %q to repository in %q %q: %s", componentName, repoPath, string(out), err)
+	}
+
+	// See if any files changed, and if so, commit and push them up to the repository
+	if out, err := e.Execute(repoPath, "git", "--no-pager", "diff", "--cached"); err != nil {
+		return fmt.Errorf("failed to check git diff in repository %q %q: %s", repoPath, string(out), err)
+	} else if string(out) != "" {
+		// Commit the changes and push
+		if out, err := e.Execute(repoPath, "git", "commit", "-m", fmt.Sprintf("Generate GitOps overlays resources for component %s and environment %s", componentName, environmentName)); err != nil {
```
johnmcollier, line 125:
```suggestion
		if out, err := e.Execute(repoPath, "git", "commit", "-m", fmt.Sprintf("Generate %s environment overlays for component %s", environmentName, componentName)); err != nil {
```""

File: gitops/gitops.go [hunk 10]
```diff
@@ -76,7 +77,52 @@ func GenerateAndPush(outputPath string, remote string, component appstudiov1alph
...
+
+	return nil
+}
+
+// GenerateOverlaysAndPush generates the overlays kustomize from App Env Snapshot Binding Spec
+func GenerateOverlaysAndPush(outputPath string, clone bool, remote string, component appstudioshared.BindingComponent, applicationName, environmentName, imageName, namespace string, e Executor, appFs afero.Afero, branch string, context string, componentGeneratedResources map[string][]string) error {
+	componentName := component.Name
+	repoPath := filepath.Join(outputPath, applicationName)
+
+	if clone {
```
johnmcollier, line 96: What's the `clone` flag for? Wouldn't we always need to clone to get the most up-to-date GitOps repository?

controllers/applicationsnapshotenvironmentbinding_controller.go
```diff
@@ -0,0 +1,241 @@
...
+			}
+			parsedURL.User = url.User(r.GitToken)
+			remoteURL = parsedURL.String()
+			clone = true
+		}
+
+		isStatusUpdated := false
+		for _, bindingStatusComponent := range appSnapshotEnvBinding.Status.Components {
+			if bindingStatusComponent.Name == componentName {
+				isStatusUpdated = true
```
yangcao77, line 186: break

controllers/applicationsnapshotenvironmentbinding_controller.go
```diff
@@ -0,0 +1,241 @@
...
+	r.SetCreateConditionAndUpdateCR(ctx, &appSnapshotEnvBinding, nil)
+
+	log.Info(fmt.Sprintf("Finished reconcile loop for %v", req.NamespacedName))
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *ApplicationSnapshotEnvironmentBindingReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		// Uncomment the following line adding a pointer to an instance of the controlled resource as an argument
```
yangcao77, line 238: cleanup the comment?

gitops/gitops.go
```diff
@@ -76,7 +77,52 @@ func GenerateAndPush(outputPath string, remote string, component appstudiov1alph
...
+			if out, err := e.Execute(repoPath, "git", "checkout", "-b", branch); err != nil {
+				return fmt.Errorf("failed to checkout branch %q in %q %q: %s", branch, repoPath, string(out), err)
+			}
+		}
+	}
+
+	// Generate the gitops resources and update the parent kustomize yaml file
+	gitopsFolder := filepath.Join(repoPath, context)
+	componentEnvOverlaysPath := filepath.Join(gitopsFolder, "components", componentName, "overlays", environmentName)
+	if err := GenerateOverlays(appFs, gitopsFolder, componentEnvOverlaysPath, component, imageName, namespace, componentGeneratedResources); err != nil {
```
yangcao77, line 112: I think you need to clean up the folder before calling `GenerateOverlays`.
Have you tried out what will happen if the overlays folder & files already exist?
IIRC, `fs.Create` will return error if the file exists. https://github.com/redhat-appstudio/application-service/blob/daa9c484120c4485af504b17059f86d518c3b887/gitops/yaml/resources.go#L58

config/samples/snapshot/snapshot.yaml
maysunfaisal: not sure how much more validation we want to do, I just did some basic sanity validation

I feel the gitops webhook should be validating these semantics or whoever is the primary controller. But your point stands, it is a valid case and i have not done it in this PR

gitops/gitops.go
maysunfaisal: so on a new reconcile call it will always be true and will be cloned and checked out

but since we in the controller we're looping the binding component(s) and generate fn is called per component, we only clone for the first component and reuse the same cloned files for the remaining components

gitops/gitops.go
maysunfaisal: Need to test this out but will defer it to a new issue potentially if it causes to break

config/kcp/apiexport_has.yaml
```diff
@@ -7,6 +7,6 @@ metadata:
   name: has
 spec:
   latestResourceSchemas:
```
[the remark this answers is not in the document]
maysunfaisal: But I wonder if this is actually HAS' responsibility? Here are my arguments:
- All the team's controllers would be eventually working on kcp in unison, so would it not be the primary controller's responsibility to generate `APIResourceSchema` for their respective APIs? Same way I would assume build-service wouldnt be generating an `APIResourceSchema` for `Component` even when they have a controller for `Component`
- When we're using HAS controllers in dev/staging; we need to explicitly install the binding, snapshot, etc CRDs; otherwise our controller will error out. So I am wondering if this is any different to how the aforementioned teams like build-service, managed-gitops, etc dont really do install `Component` CRDs even if they have a controller for it 

🤔

controllers/applicationsnapshotenvironmentbinding_controller_conditions.go
```diff
@@ -0,0 +1,50 @@
...
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func (r *ApplicationSnapshotEnvironmentBindingReconciler) SetCreateConditionAndUpdateCR(ctx context.Context, appSnapshotEnvBinding *appstudioshared.ApplicationSnapshotEnvironmentBinding, createError error) {
+	log := r.Log.WithValues("ApplicationSnapshotEnvironmentBinding", appSnapshotEnvBinding.Name)
+
+	if createError == nil {
+		meta.SetStatusCondition(&appSnapshotEnvBinding.Status.GitOpsRepoConditions, metav1.Condition{
+			Type:    "Sync",
```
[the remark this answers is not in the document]
johnmcollier: `GitOpsResourcesGenerated`

config/kcp/apiexport_has.yaml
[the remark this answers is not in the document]
johnmcollier: You're correct that it's not our responsibility, but I was making a note just that this will complicate testing on KCP until `ThingPermissionClaim` support is in KCP, and the APIResourceSchemas / APIExports for the Snapshot & Binding CRDs are generated by GitOps.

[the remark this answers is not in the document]
johnmcollier: Until then, we'll need to manually generate the APIResourceSchemas and add it to our APIExports before testing.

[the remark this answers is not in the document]
maysunfaisal: ok, do we just copy paste the CR spec in the field there for these new CRs? 🤔

controllers/applicationsnapshotenvironmentbinding_controller_conditions.go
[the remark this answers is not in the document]
maysunfaisal: I'll change to `GitOpsResourcesGenerated`

controllers/applicationsnapshotenvironmentbinding_controller_conditions.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"context"
+	"fmt"
+
+	appstudioshared "github.com/redhat-appstudio/managed-gitops/appstudio-shared/apis/appstudio.redhat.com/v1alpha1"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func (r *ApplicationSnapshotEnvironmentBindingReconciler) SetCreateConditionAndUpdateCR(ctx context.Context, appSnapshotEnvBinding *appstudioshared.ApplicationSnapshotEnvironmentBinding, createError error) {
```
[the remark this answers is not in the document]
maysunfaisal: Im updating to `SetConditionAndUpdateCR` since I am changing to `GitOpsResourcesGenerated`

controllers/applicationsnapshotenvironmentbinding_controller.go
maysunfaisal: this was tied to the clone logic, i have removed it in my recent commit

Makefile
```diff
@@ -275,7 +275,8 @@ apply-crds:
 	kubectl apply \
 	-f $(APPLICATIONS_CRD) \
 	-f $(COMPONENT_DETECTION_QUERIES_CRD) \
-	-f $(COMPONENT_CRD)
+	-f $(COMPONENT_CRD) \
+	-f https://raw.githubusercontent.com/redhat-appstudio/managed-gitops/main/appstudio-shared/manifests/appstudio-shared-customresourcedefinitions.yaml
```
johnmcollier, line 279: Should the `install` makefile target be updated to apply the shared CRDs? Then we can cut out the section in the readme about installing the crd.

maysunfaisal: updated

pkg/util/util.go
```diff
@@ -49,6 +50,36 @@ func IsExist(path string) (bool, error) {
 	return false, err
 }
 
+// ProcessGitOpsStatus processes the GitOps status and returns the remote url, branch, context and the error
+func ProcessGitOpsStatus(gitopsStatus appstudiov1alpha1.GitOpsStatus, gitToken string) (string, string, string, error) {
```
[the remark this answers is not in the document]
johnmcollier: I guess what I'm getting at is, wouldn't be it be useful for the GitOpsStatus to say that the branch is `main` or that the context is `/`?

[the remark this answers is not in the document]
maysunfaisal: discussed in the meeting, we will keep it as it is and open an issue to track this as we would need to up date some of the logic. For instance, we cannot assume a repo's default branch is `main`

johnmcollier (approved):

PR #136, opened 2022-06-23
PLNSRVCE-430: Initial integration of Pipelines-as-Code

Michkov:
By setting envvar PAC_ENABLED=1 is Repository CR of Pipelines-as-Code
created instead of tekton triggers.


elsony (approved):
/lgtm

PR #139, opened 2022-06-27
[DEVHAS-106] Add SkipGitOpsResourceGeneration field

johnmcollier:
Fixes https://issues.redhat.com/projects/DEVHAS/issues/DEVHAS-106

Adds a new field to the Component spec (`skipGitOpsResourceGeneration`) that allows GitOps resource generation to be disabled when it is set to true. If the field is unset, or set to false, the GitOps resources will still be generated, as-is today.

I've also added a new status condition to the CR, `GitOpsResourcesGenerated`, which denotes the last time the GitOps resources for the component were generated by the controller. I've also updated the tests accordingly to check for this condition and to take it into account when retrieving the CR's conditions


controllers/component_controller_test.go
```diff
@@ -783,12 +789,18 @@ var _ = Describe("Component controller", func() {
 			createdHasComp := &appstudiov1alpha1.Component{}
 			Eventually(func() bool {
 				k8sClient.Get(context.Background(), hasCompLookupKey, createdHasComp)
-				return len(createdHasComp.Status.Conditions) > 0 && createdHasComp.Status.Conditions[len(createdHasComp.Status.Conditions)-1].Status == metav1.ConditionFalse
+				return len(createdHasComp.Status.Conditions) > 1 && createdHasComp.Status.Conditions[len(createdHasComp.Status.Conditions)-1].Status == metav1.ConditionFalse
```
yangcao77, line 792: should here check `Conditions[len(createdHasComp.Status.Conditions)-2].Status`?

yangcao77 (changes requested):
missing unit tests for `skipGitOpsResourceGeneration` flag being added.

controllers/component_controller_test.go
johnmcollier: Shouldn't be needed. Because we add the `Created` status condition _after_ the gitops status condition, waiting til the `Created` condition is added is sufficient to "verify" that the reconcile has concluded.

We also check `Conditions[len(createdHasComp.Status.Conditions)-2].Status` a few lines down below.

johnmcollier:
@maysunfaisal @yangcao77 rebased

api/v1alpha1/component_types.go
```diff
@@ -138,6 +141,8 @@ type GitOpsStatus struct {
 
 	// Context is the path within the gitops repository used for the gitops resources
 	Context string `json:"context,omitempty"`
+
+	ResourceGenerationSkipped bool `json:"resourceGenerationSkipped,omitempty"`
```
maysunfaisal, line 145: probably add a quick desc

maysunfaisal: since its in types.go

controllers/applicationsnapshotenvironmentbinding_controller_test.go
```diff
@@ -1266,6 +1266,157 @@ var _ = Describe("ApplicationSnapshotEnvironmentBinding controller", func() {
 		})
 	})
 
+	Context("Create ApplicationSnapshotEnvironmentBinding with some component configurations", func() {
+		It("Should generate gitops overlays successfully", func() {
```
maysunfaisal, line 1270: update the test name pls

controllers/component_controller.go
```diff
@@ -374,23 +379,30 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+			component.Status.GitOps.ResourceGenerationSkipped = skipGitOpsGeneration
 			yamlHASCompData, err := yaml.Marshal(hasCompDevfileData)
 			if err != nil {
 				log.Error(err, fmt.Sprintf("Unable to marshall the Component devfile, exiting reconcile loop %v", req.NamespacedName))
 				r.SetUpdateConditionAndUpdateCR(ctx, &component, err)
 				return ctrl.Result{}, nil
 			}
 
-			// Generate and push the gitops resources
+			// Generate and push the gitops resources, if necessary.
```
maysunfaisal, line 394: Just a Q out of curiosity. What happens in this case?

1. skipGitOpsGeneration is false during Component create, gitops resources generated
2. skipGitOpsGeneration is updated to true during Component update, gitops resources not generated/updated

So the gitops resources generated during step 1 will still remain and be possibly be out of date after step 2. Should they remain as is in the gitops repo?

controllers/component_controller_conditions.go
```diff
@@ -75,3 +75,28 @@ func (r *ComponentReconciler) SetUpdateConditionAndUpdateCR(ctx context.Context,
...
+			Type:    "GitOpsResourcesGenerated",
+			Status:  metav1.ConditionTrue,
+			Reason:  "OK",
+			Message: "GitOps resource generated successfully",
+		})
+	} else {
+		meta.SetStatusCondition(&component.Status.Conditions, metav1.Condition{
+			Type:    "GitOpsResourcesGenerated",
+			Status:  metav1.ConditionFalse,
+			Reason:  "GenerateError",
```
maysunfaisal, line 93: can you also update the file `controllers/applicationsnapshotenvironmentbinding_controller_conditions.go` with this `Reason`

maysunfaisal: or if we want to do `OK` and `Error`, update this instead

controllers/component_controller.go
johnmcollier: Yes, they should remain as-is on the gitops repo in that scenario.

api/v1alpha1/component_types.go
johnmcollier: Oops, fixed.

controllers/component_controller_conditions.go
johnmcollier: Updated to use GenerateError.

johnmcollier:
@maysunfaisal Review comments addressed.

maysunfaisal (approved):

yangcao77 (approved):
lgtm

PR #144, opened 2022-07-05
[DEVHAS-134] Sanitize Git error messages

johnmcollier:
Fixes https://issues.redhat.com/browse/DEVHAS-134

Adds a new util function (+ tests) for sanitizing error messages. Can expand in the future as needed. Updates the Component controller to call this function before reporting errors from the gitops library.

### DIFF
--- a/controllers/component_controller.go
+++ b/controllers/component_controller.go
@@ -475,8 +475,9 @@ func (r *ComponentReconciler) generateGitops(ctx context.Context, component *app
 
 	err = gitops.GenerateAndPush(tempDir, remoteURL, *component, r.Executor, r.AppFS, gitOpsBranch, gitOpsContext, gitopsConfig)
 	if err != nil {
-		log.Error(err, "unable to generate gitops resources due to error")
-		return err
+		gitOpsErr := util.SanitizeErrorMessage(err)
+		log.Error(gitOpsErr, "unable to generate gitops resources due to error")
+		return gitOpsErr
 	}
 
 	// Remove the temp folder that was created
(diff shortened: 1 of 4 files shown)


controllers/component_controller.go
```diff
@@ -475,8 +475,9 @@ func (r *ComponentReconciler) generateGitops(ctx context.Context, component *app
 
 	err = gitops.GenerateAndPush(tempDir, remoteURL, *component, r.Executor, r.AppFS, gitOpsBranch, gitOpsContext, gitopsConfig)
 	if err != nil {
-		log.Error(err, "unable to generate gitops resources due to error")
-		return err
+		gitOpsErr := util.SanitizeErrorMessage(err)
```
maysunfaisal, line 478: needs to be also called on `RemoveAndPush()`

johnmcollier: Good catch

johnmcollier: Fixed

maysunfaisal:
@johnmcollier I wonder if we should as part of a diff issue sanitize this more broadly.. For instance, 

1. directly call the util fn in *_controller_conditions.go
2. create a log wrapper, to filter out any possible future confidential info

this way we can maintain a list of regex(s) and not worry about where to call it 🤔

johnmcollier:
@maysunfaisal Yes, agreed.

I do some concerns around the overhead of too many regex checks on _every_ log statement, but we can look into that

maysunfaisal (approved):

PR #145, opened 2022-07-06
Use the Component revision

jduimovich:
This PR contains a fix for Components where the default branch is other than `main` and is specified in the the Component Spec as part of the source git reference

The repository https://github.com/Apicurio/apicurio-registry uses `master` instead of main but the reference in the component is ignored when a build is created and the build fails. This PR adds a check to setthe revision into the pipeline run if it is set in the component. 

Example (note the `revision: master`) 
```
kind: Component
metadata:
  name: service-registry
spec:
  componentName: service-registry
  application: service-registry
  containerImage: quay.io/jduimovich0/service-registry
  source:
    git:
      url: https://github.com/Apicurio/apicurio-registry
      revision: master
      devfileUrl: https://raw.githubusercontent.com/jduimovich/appstudio-e2e-demos/main/demos/service-registry/devfiles/devfile.yaml
  build:
    containerImage: quay.io/jduimovich0/service-registry:1
```
 
 The application definition for the above component.  It will fail to build as the default will checkout `main`, instead of `master`. 
```
kind: Application
metadata:
  name: service-registry 
spec:
  description: "service-registry built in App Studio"
  displayName: service-registry
```

To test, rebuild both application-service and build-service as the code is shared in both and create a Component using the above as an example build.

### DIFF
--- a/gitops/generate_build.go
+++ b/gitops/generate_build.go
@@ -234,7 +234,7 @@ func normalizeOutputImageURL(outputImage string) string {
 // is that the git revision appended to the output image tag in case of webhook build.
 func getParamsForComponentBuild(component appstudiov1alpha1.Component, isInitialBuild bool) ([]tektonapi.Param, error) {
 	sourceCode := component.Spec.Source.GitSource.URL
-	revision := component.Spec.Source.GitSource.Revision 
+	revision := component.Spec.Source.GitSource.Revision
 	outputImage := component.Spec.ContainerImage
 	var err error
 
@@ -265,14 +265,14 @@ func getParamsForComponentBuild(component appstudiov1alpha1.Component, isInitial
 	}
 	// if revision is specified in the component
 	// use it in the parms to the Pipeline Run
-	if revision != "" { 
+	if revision != "" {
 		params = append(params, tektonapi.Param{
 			Name: "revision",
 			Value: tektonapi.ArrayOrString{
 				Type:      tektonapi.ParamTypeString,
 				StringVal: revision,
 			},
-		}) 
+		})
 	}
 	// Analyze component model for additional parameters
 	if componentDevfileData, err := devfile.ParseDevfileModel(component.Status.Devfile); err == nil {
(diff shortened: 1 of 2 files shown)


gabemontero:
a hacbs / jvm build service need for sure, given our long term goal of building service registry end to end from appstudio invoking jvm build service via creation of a component

thanks @jduimovich 

@sbose78 @Michkov @stuartwdouglas (for when they return from PTO)

gabemontero (commented):
you have time to write a unit test @jduimovich or shall @mmorhun or myself add one after this gets merged ?

johnmcollier (commented):
Can you add a test case to [TestGetParamsForComponentBuild](https://github.com/redhat-appstudio/application-service/blob/99bd9270ab7a72b7ff73c5d745b06943bc6d0690/gitops/generate_build_test.go#L652) where the Git Source revision is set to something?

jduimovich:
I added a test case for the revision param use case as well and fixing the formatting (extra spaces)

gabemontero:
thanks @jduimovich 

the `kcp` submodule you added appears to be causing problems with the image build:

`{"command":"PWD=/go/src/github.com/redhat-appstudio/application-service  git submodule update --init --recursive","component":"clonerefs","error":"exit status 1","file":"k8s.io/test-infra/prow/pod-utils/clone/clone.go:83","func":"k8s.io/test-infra/prow/pod-utils/clone.Run.func2","level":"info","msg":"Ran command","output":"No submodule mapping found in .gitmodules for path 'kcp'\n","severity":"info","time":"2022-07-07T03:30:28Z"}`

maybe try to take that on in a separate PR (unless the resolution here is obvious to you) ?

jduimovich:
@gabemontero  --- yes, adding kcp was an accident, when the `make test`  failed it prints the instruction to build kcp. I thought I had checked the commits and had removed it pre-pr

gabemontero:
/lgtm

/assign @johnmcollier 

I think we are good to go - just need the approve label from @johnmcollier

PR #150, opened 2022-07-13
[DEVHAS-127] Support KCP virtual workspaces

johnmcollier:
This PR adds support for KCP virtual workspaces to HAS. 

When running in a KCP workspace, HAS can be started with the `--api-export-name <export>` flag, and HAS will retrieve the corresponding virtual workspace, and initialize the operator with it. Furthermore, Kubernetes client requests in each of the 4 controllers that HAS manages have been updated to be multicluster aware. 

In all, this PR does the following:
- Pulls in the [kcp-dev/controller-runtime](https://github.com/kcp-dev/controller-runtime) fork, and replaces the existing controller-runtime library with it
- Updates the operator initialization code in `main.go`:
   - If running on a KCP environment: retrieves the virtual workspace from the passed in `APIExport` for the HAS CRDs, and initializes the operator with the virtual workspace
   - If running on a non-KCP (i.e. regular Kube) environment, the operator is initialized as usual
- Updates the log and context in each controller to include the logicalcluster for the given reconciliation request.
   - The logicalcluster value will be blank when on non-KCP
   
## How to Test

1) Start KCP:

   a) Git clone https://github.com/kcp-dev/kcp
   b) Run `make install`
   c) Run `kcp start` to start the KCP instance

2) Create a workspace and deploy APIResoruceSchema & APIExport resources:
   
   a) Open a new terminal window and export the `KUBECONFIG` variable to `$KCP/.kcp/auth.kubeadmin`
   b) Run `kubectl ws list`. If no workspace is present, run `kubectl ws create users`
   c) Run `kubectl ws use users` 
   d) Create the `APIResoruceSchema` & `APIExport` resources with: `kubectl create -f ~/kcp/application-service/config/kcp/apiresourceschema_has.yaml` and `kubectl create -f ~/kcp/application-service/config/kcp/apiexport_has.yaml`

3) Create an APIBinding resource:
   ```bash
   echo "apiVersion: apis.kcp.dev/v1alpha1
   kind: APIBinding
   metadata:
     name: has
   spec:
     reference:
       workspace:
         path: root:users
         exportName: has" | kubectl create -f -

   ```
   
4) Start HAS: `make bin` followed by `./bin/manager --api-export-name has`
5) Test HAS:
   a) Open a new terminal window and export the `KUBECONFIG` variable to `$KCP/.kcp/auth.kubeadmin`
   b) Create Application/Component/CDQ resources as needed


controllers/componentdetectionquery_controller.go
```diff
@@ -64,7 +66,8 @@ type ComponentDetectionQueryReconciler struct {
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.9.2/pkg/reconcile
 func (r *ComponentDetectionQueryReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	log := r.Log.WithValues("ComponentDetectionQuery", req.NamespacedName)
+	log := r.Log.WithValues("ComponentDetectionQuery", req.NamespacedName).WithValues("clusterName", req.ClusterName)
```
maysunfaisal, line 69: wonder if you only want to add the clustername in the controller files? Other files that use logs will not have this value..

johnmcollier: Good catch. We should have it for all of the cases where we create a logger. Will need to pass in the request struct into those functions

johnmcollier: @maysunfaisal Updated.

johnbieren:
@johnmcollier just for context since we are doing a similar thing in the release-service, can you say why the logic for `apiExportName == ""` isn't needed in `main.go`? https://github.com/kcp-dev/controller-runtime-example/blob/main/main.go#L176..L193 for reference. Here, we just have lines 177 and 178, so I guess `apiExportName` will not be the empty string

johnmcollier:
@johnbieren Good question. The function I was using in main.go was a little older, and didn't have those changes, but since I was always passing in the apiexport name when testing on KCP, the missing lines for `apiExportName == ""` weren't needed.

However, they're good to have, so I've rebased to add those changes in.

maysunfaisal (approved):

PR #151, opened 2022-07-13
[DEVHAS-109] custom kustomization patch

yangcao77:
This PR includes the change for DEVHAS109- phase 2 in https://docs.google.com/document/d/10VEVAG2H2LpntgUV6y_ETNE1ruVuZJ2Xue1nIhlMrHI/edit#heading=h.9ck1iviree1k

With this change, custom patches will be preserved in overlays folders, and also in the kustomization file, upon any changes. 
The PR also resolves the import issue when bring user's own gitops repo, removed the `rm -rf` command when generate component to preserve user's component content

To test:
1.create `application-sample` with gitops url: https://github.com/yangcao77/petclinic-application-service-system-predict-indicate
2. create `component-sample` & `component-sample2`
3. create `snapshot` and `binding`
4. update `binding` 
5. check the overlays folder in gitops repo, the test1.yaml and test2.yaml should still persist in both overlays folder and in kustomization patches field.


gitops/yaml/resources.go
```diff
@@ -75,3 +75,20 @@ func MarshalOutput(out io.Writer, output interface{}) error {
...
+func UnMarshalItemFromFile(fs afero.Fs, filename string, item interface{}) error {
+	content, err := afero.ReadFile(fs, filename)
+	if err != nil {
+		return fmt.Errorf("failed to read from file %s: %v", filename, err)
+	}
+
+	err = yaml.Unmarshal(content, item)
+	if err != nil {
+		return fmt.Errorf("failed to unmarshal data: %v", err)
+	}
```
maysunfaisal, line 91: we can get test coverage % for this block by just trying to mimic a test case from go pkg:
Screen Shot 2022-07-18 at 4 59 37 PM

gitops/generate.go
```diff
@@ -90,6 +91,23 @@ func Generate(fs afero.Afero, gitOpsFolder string, outputFolder string, componen
 
 // GenerateOverlays generates the overlays dir from a given BindingComponent
 func GenerateOverlays(fs afero.Afero, gitOpsFolder string, outputFolder string, component appstudioshared.BindingComponent, imageName, namespace string, componentGeneratedResources map[string][]string) error {
+	kustomizeFileExist, err := fs.Exists(filepath.Join(outputFolder, kustomizeFileName))
```
maysunfaisal, line 94: just do a quick search and see if you can hit the error cases using a different fs? Probably just check the module tests to see if we can do it as well

gitops/resources/kustomization_test.go
```diff
@@ -67,3 +67,13 @@ func Test_AddResource_sorts_elements(t *testing.T) {
 		t.Fatalf("failed to sort resources:\n%s", diff)
 	}
 }
+
+func Test_CompareDifferenceAndAddCustomizedPatches(t *testing.T) {
+	k := Kustomization{}
+	original := []string{"tesing.yaml", "testing2.yaml", "custom.yaml", "custom2.yaml"}
```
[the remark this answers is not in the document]
yangcao77: updated

gitops/yaml/resources.go
yangcao77: added test coverage for the unmarshal data and also marshal data failing case

gitops/generate.go
yangcao77: took a quick look, a different fs does not return any error, it will only return error from [os.stat](https://cs.opensource.google/go/go/+/refs/tags/go1.18.4:src/os/stat_unix.go;l=20)
and only if `File` is nil or did not setup correctly. In our code, we do not touch down to lowlevel `File` object. 
and `GenerateOverlays` has no test for it, it's being tested as part of `GenerateAndPushOverlays` test, if the path is invalid, it will be found out and returned before it reaches to this point

gitops/gitops.go
```diff
@@ -57,13 +59,7 @@ func GenerateAndPush(outputPath string, remote string, component appstudiov1alph
 		}
 	}
 
-	if out, err := e.Execute(repoPath, "rm", "-rf", filepath.Join("components", componentName)); err != nil {
```
johnmcollier, line 60: If resources change between generations, and the subsequent gitops generation don't include a resource that was present in the previous generation, then the old resources would be left behind and not removed (see below for an example)

E.g., right now, when we generate gitops resources for a given component, if a component specifies a targetPort, we'll include a route in the resources that we generate. But if a user were to remove that targetPort, the next generation wouldn't include the route resource, but since we're no blowing away the base folder.

We could either blow away the base folder on each generation (but this would cause issues with my other suggestion) or we could:
1) Read the old kustomize.yaml
2) Then parse the entries under `base` and delete each file before generation 

It's a bit of an edge case, but I think it's something we need to handle, especially once we generate gitops resources from the Kubernetes resources in the devfile.

maysunfaisal: We are doing a git diff right? So I think it should catch the subsequent removal of targetPort, since we will not generate the route and there would be a git diff 🤔

johnmcollier: Hmm? But in this case, the route file won't be touched, since it's not one of the files we're generating, so it won't be included in the git diff

gitops/generate.go
```diff
@@ -90,6 +91,23 @@ func Generate(fs afero.Afero, gitOpsFolder string, outputFolder string, componen
 
 // GenerateOverlays generates the overlays dir from a given BindingComponent
 func GenerateOverlays(fs afero.Afero, gitOpsFolder string, outputFolder string, component appstudioshared.BindingComponent, imageName, namespace string, componentGeneratedResources map[string][]string) error {
```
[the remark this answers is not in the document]
johnmcollier: I guess what I mean is, why aren't custom patches in the base folder supported?

gitops/gitops.go
yangcao77: git diff might only catch what has been editted. so John was right, the route.yaml wouldn't be removed in this case. 
I'm wondering if that's something affecting the behavior? removing the targetPort will remove the `route.yaml` from the kustomization file, so should not affect anything right? just a dangling `route.yaml` left in the gitops folder.

gitops/generate.go
johnmcollier: Should we have a test case for GenerateOverlays then? Would be good to have a test case that validates the contents of the overlays we're generating (similar to `TestGenerateParentKustomize` in https://github.com/redhat-appstudio/application-service/blob/729bfb4664b9a6cd5821a90c65fc655176fd5b87/gitops/generate_test.go#L575)

gitops/gitops.go
maysunfaisal: so would it be an issue if there is a route.yaml left behind if its not mentioned in the kustomize.yaml 🤔 

the other alternative is to read the kustomize file and cross check with the files in the dir and remove any unnecessary ones but if my earlier comment makes sense ie kustomize.yaml doesnt have an entry then we can skip this..

gitops/generate.go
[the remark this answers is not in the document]
yangcao77: emm. I also have the same question when I was checking the design doc. looks like only overlays folder is expected to have custom patches provided. @maysunfaisal do you guys have any discussion on that with HACBS team?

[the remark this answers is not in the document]
johnmcollier: Since it's part of the phase 2 design, I won't block the PR on leaving it out, but it would be worthwhile having a discussion on it 🤔

gitops/gitops.go
yangcao77: emm. I think it also depends this question John raised https://github.com/redhat-appstudio/application-service/pull/151#discussion_r924847171
if we allow custom patches under base folder, then we should not blow the folder or delete any files. (even it's not included in the `kustomization.yaml` file, might be a custom file temporarily removed by user, should not delete the source file). If we do not allow custom patches under base folder, then I think we can assume the base folder only contains generated files, so base folder can be safely deleted.

gitops/generate.go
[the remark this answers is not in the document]
maysunfaisal: Yeah I wasnt aware of base dir having custom patches. It was more targeted towards an environment. 

I guess if a user really want to have some sort of base patch for all the environment, then they should put those configurations in the component level from the UI.. and it will get reflected in the base dir generation

gitops/generate.go
yangcao77: I've added a `TestGenerateOverlays` with unittests. I'm not able to cover the `os.stat` failure case tho, but it should cover other unreached lines and increase coverage.

johnmcollier: It'll be tricky to test for os.stat failures, so I think this is fine.

gitops/gitops.go
johnmcollier: Let's just blow away the base dir for now? And if we decide to allow custom patches under the base folder late, we'll handle deletion like I suggested in https://github.com/redhat-appstudio/application-service/pull/151#discussion_r924835458

gitops/generate.go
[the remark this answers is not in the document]
johnmcollier: Let's leave as is for now, and can discuss later if we need/want to support patches under the base folder?

johnmcollier (approved):

PR #153, opened 2022-07-18
Set ArgoCD prune option to false for 'appstudio' pvc

mmorhun:
This prevents `appstudio` volume from switching into deleting phase if a `Component` gets deleted.

### DIFF
--- a/gitops/generate_build.go
+++ b/gitops/generate_build.go
@@ -342,6 +342,9 @@ func getBuildCommonLabelsForComponent(component *appstudiov1alpha1.Component) ma
 func GenerateCommonStorage(component appstudiov1alpha1.Component, name string) *corev1.PersistentVolumeClaim {
 	fsMode := corev1.PersistentVolumeFilesystem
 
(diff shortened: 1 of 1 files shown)


gitops/generate_build.go
```diff
@@ -342,6 +342,9 @@ func getBuildCommonLabelsForComponent(component *appstudiov1alpha1.Component) ma
 func GenerateCommonStorage(component appstudiov1alpha1.Component, name string) *corev1.PersistentVolumeClaim {
 	fsMode := corev1.PersistentVolumeFilesystem
 
+	annotations := getBuildCommonLabelsForComponent(&component)
+	annotations["argocd.argoproj.io/sync-options: Prune"] = "false"
```
Michkov, line 346: Key is `argocd.argoproj.io/sync-options` and value is `Prune=false`, so it should be: `annotations["argocd.argoproj.io/sync-options"] = "Prune=false"`

mmorhun:
This solution will not work in case of deletion of ArgoCD app object.

PR #156, opened 2022-07-26
Added ppc64le support

mayurwaghmode:
This PR will add multi-architecture support to the application-service operator image


johnmcollier (commented):
Unfortunately, the github workflow that we use on PR merge is redundant, and has been replaced by this Tekton job: https://github.com/redhat-appstudio/application-service/blob/main/.tekton/push.yaml

I'm not aware how to publish ppc64le images with it, however.

johnmcollier (approved):

PR #158, opened 2022-07-27
[DEVHAS-123] Update Dockerfile URL for relative URI

maysunfaisal:

This PR updates the CDQ status for detected components. It includes the Dockerfile URL for
- any components detected using Alizer
- any devfile that has a Dockerfile URL reference

### DIFF
--- a/pkg/devfile/detect.go
+++ b/pkg/devfile/detect.go
@@ -140,33 +140,37 @@ func AnalyzePath(a Alizer, localpath, context, devfileRegistryURL string, devfil
 		if dockerfileUri, err := SearchForDockerfile(devfile); err != nil {
 			return err
 		} else if len(dockerfileUri) > 0 {
+			dockerfileContextMapFromRepo[context] = path.Join(context, dockerfileUri)
 			isDockerfilePresent = true
 		}
 	}
 
 	if !isDockerfilePresent {
 		// if we didnt find any devfile/dockerfile upto our desired depth, then use alizer
-		devfile, detectedDevfileEndpoint, detectedSampleName, err := AnalyzeAndDetectDevfile(a, localpath, devfileRegistryURL)
+		detectedDevfile, detectedDevfileEndpoint, detectedSampleName, err := AnalyzeAndDetectDevfile(a, localpath, devfileRegistryURL)
 		if err != nil {
 			if _, ok := err.(*NoDevfileFound); !ok {
 				return err
 			}
 		}
 
-		if !isDevfilePresent && len(devfile) > 0 {
-			// If a devfile is not present at this stage, just update devfileMapFromRepo and devfilesURLMapFromRepo
-			// Dockerfile is not needed because all the devfile registry samples will have a Dockerfile entry
-			devfileMapFromRepo[context] = devfile
-			devfilesURLMapFromRepo[context] = detectedDevfileEndpoint
-		} else if isDevfilePresent && len(devfile) > 0 {
-			// If a devfile is present but we could not determine a dockerfile, then update dockerfileContextMapFromRepo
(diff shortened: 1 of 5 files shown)


yangcao77 (approved):
changes looks good to me

controllers/component_controller.go
```diff
@@ -393,7 +393,7 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 
 			// Generate and push the gitops resources, if necessary.
 			component.Status.ContainerImage = component.Spec.ContainerImage
-      if !component.Spec.SkipGitOpsResourceGeneration {
+			if !component.Spec.SkipGitOpsResourceGeneration {
```
johnmcollier, line 396: I have no idea how this happened lol, cause go fmt should've caught this.

pkg/devfile/detect.go
```diff
@@ -140,33 +140,37 @@ func AnalyzePath(a Alizer, localpath, context, devfileRegistryURL string, devfil
...
 		}
 
-		if !isDevfilePresent && len(devfile) > 0 {
-			// If a devfile is not present at this stage, just update devfileMapFromRepo and devfilesURLMapFromRepo
-			// Dockerfile is not needed because all the devfile registry samples will have a Dockerfile entry
-			devfileMapFromRepo[context] = devfile
-			devfilesURLMapFromRepo[context] = detectedDevfileEndpoint
-		} else if isDevfilePresent && len(devfile) > 0 {
-			// If a devfile is present but we could not determine a dockerfile, then update dockerfileContextMapFromRepo
+		if len(detectedDevfile) > 0 {
```
johnmcollier, line 157: Nit: Couldn't we still keep it as `if !isDevfilePresent && len(detectedDevfile) > 0`?

pkg/devfile/detect.go
```diff
@@ -140,33 +140,37 @@ func AnalyzePath(a Alizer, localpath, context, devfileRegistryURL string, devfil
 		if dockerfileUri, err := SearchForDockerfile(devfile); err != nil {
 			return err
 		} else if len(dockerfileUri) > 0 {
+			dockerfileContextMapFromRepo[context] = path.Join(context, dockerfileUri)
```
johnmcollier, line 143: Dumb question: Can `dockerfileUri` be a URL in any case? E.g. some scenario where we wouldn't want to automatically append the context?

yangcao77: you are right. in our api reference: `URI Reference of a Dockerfile. It can be a full URL or a relative URI from the current devfile as the base URI.`

pkg/devfile/detect.go
maysunfaisal: there is also some code after `if !isDevfilePresent {}` block which is executed for when both a devfile is present and not present:

1. Devfile NOT Present: Set the Alizer detected Devfile
2. Devfile NOT Present & Devfile Present: Set the Dockerfile full URL

maysunfaisal:
```
if len(detectedDevfile) > 0 {
			if !isDevfilePresent {
				// Save the Alizer detected Devfile
			}
			
                         // Save the Dockerfile full URL
			isDockerfilePresent = true
		}
```

pkg/devfile/detect.go
maysunfaisal: true, my bad

pkg/devfile/detect.go
johnmcollier: I clearly cannot read lol. Okay that makes sense, dunno why I missed the code underneath.

johnmcollier (approved):

johnmcollier (approved):

PR #164, opened 2022-08-15
[DEVHAS-151] Add PermissionClaims for HAS ApiExports

johnmcollier:
Updates the KCP API generator script to include Permission Claims for the non-KCP dependencies that HAS has (secrets, routes, configmaps). Also updated the APIExport for HAS to the preferred format (`appstudio-identity-<component-name>`)

### DIFF
--- a/config/kcp/apiexport_has.yaml
+++ b/config/kcp/apiexport_has.yaml
@@ -4,8 +4,18 @@
 apiVersion: apis.kcp.dev/v1alpha1
 kind: APIExport
 metadata:
-  name: has
+  name: apiexport-identity-has
 spec:
+  permissionClaims:
+  - group: ""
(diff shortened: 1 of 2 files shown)


config/kcp/apiexport_has.yaml
```diff
@@ -4,8 +4,18 @@
 apiVersion: apis.kcp.dev/v1alpha1
 kind: APIExport
 metadata:
-  name: has
+  name: apiexport-identity-has
 spec:
+  permissionClaims:
+  - group: ""
+    resource: "secrets"
+    verbs: ["get", "list", "watch"]
```
mmorhun, line 12: AFAIK `verbs` is not in the spec of `APIExport` CRD

PR #165, opened 2022-08-16
Use Go 1.18 in GitHub workflow

johnmcollier:
KCP now requires Go 1.18 to build now, so need to update our PR workflow to use it.


elsony (approved):
/lgtm

PR #166, opened 2022-08-22
[DEVHAS-148] Add GitOps commit ID to Component status

johnmcollier:
Part 1 of https://issues.redhat.com/browse/DEVHAS-148

This PR adds a new field to the status of Component resources, `commitID`, which corresponds to the most recent commitID in the GitOps repository for that given Component. To retrieve the commitID, I added a function `GetCommitSHAFromRepo`, which calls `git rev-parse HEAD` on the given repository folder. (Note: There are multiple approaches to get the commit ID, but this approach was the most straightforward and testable). 

For the controller unit tests, which do not use 'real' GitOps repositories (which means there is no commit SHA to retrieve), I updated the mock executor that we use to return a dummy commit SHA when `git rev-parse HEAD` is called.

### DIFF
--- a/api/v1alpha1/component_types.go
+++ b/api/v1alpha1/component_types.go
@@ -141,6 +141,9 @@ type GitOpsStatus struct {
 
 	// ResourceGenerationSkipped is whether or not GitOps resource generation was skipped for the component
 	ResourceGenerationSkipped bool `json:"resourceGenerationSkipped,omitempty"`
+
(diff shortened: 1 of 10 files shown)


johnmcollier:
@yangcao77 Updated to change all references of Commit SHA to Commit ID for consistency 👍

yangcao77 (approved):
changes looks good to me

PR #169, opened 2022-08-23
[DEVHAS-143] unique name in cdq

yangcao77:
This PR  generates unique component name in cdq spec

1. get the repo name as component name
2. append folder name if contains multiple component
3. validate the name for dns1123 label format. remove invalid characters, convert to all lower case, add prefix `comp-` if the  name is all numeric, restrict the length to be within 63 chars. 
4. check for existing component. if conflict, append random 4 chars at the end


controllers/componentdetectionquery_controller_conditions.go
```diff
@@ -61,9 +69,65 @@ func (r *ComponentDetectionQueryReconciler) SetCompleteConditionAndUpdateCR(ctx
...
+		name = name[0:58]
+	}
+
+	// get hc
+	hc := &appstudiov1alpha1.Component{}
+	namespacedName := types.NamespacedName{
+		Name:      name,
+		Namespace: namespace,
+	}
+	newErr := client.Get(ctx, namespacedName, hc)
```
johnmcollier, line 125: I'm not crazy about us doing a Kubernetes API request each time we need to sanitize the name, which will be for the N components returned by a CDQ. There's already some concerns around the response time for CDQs, and this will likely slow it down further.

Could we just always add on the 4 random characters, and skip the calls to the Kubernetes API?

controllers/componentdetectionquery_controller_conditions.go
```diff
@@ -61,9 +69,65 @@ func (r *ComponentDetectionQueryReconciler) SetCompleteConditionAndUpdateCR(ctx
 			Message: fmt.Sprintf("ComponentDetectionQuery failed: %v", completeError),
 		})
 	}
+	updateComponentName(ctx, componentDetectionQuery, r.Client)
```
maysunfaisal, line 72: Why not call this in `controllers/update.go` func `updateComponentStub()`?

1. At this point we have two name generation, we do it once in `updateComponentStub()` and then again in `SetCompleteConditionAndUpdateCR()`
2. Seems like we want to keep conditions file simple; its for only setting the status conditions

johnmcollier (changes requested):
One primary concern

controllers/componentdetectionquery_controller_conditions.go
maysunfaisal:
```componentName := "component"
		if len(devfileMetadata.Name) > 0 {
			// use regex to see if metadata.name can be used as component name, since this will be used in kube resources
			if matched := util.CheckWithRegex("^[a-z]([-a-z0-9]*[a-z0-9])?", devfileMetadata.Name); matched {
				componentName = strings.ToLower(devfileMetadata.Name)
			}
		}
		componentName = checkComponentName(componentName, componentDetectionQuery.Status.ComponentDetected)
```

and much further down

```
componentName := "component"
		componentName = checkComponentName(componentName, componentDetectionQuery.Status.ComponentDetected)
```

controllers/componentdetectionquery_controller_conditions.go
yangcao77: I raised this question during our meeting, but last time Elson still wants to check for conflict instead of  always adding random character. 
I do agree the current CDQ response time is already very long when using HAC UI, and it's reasonable to avoid unnecessary slowing down. but the actual response time when running HAS through command line is not that bad.

We can discuss further when Elson comes back.

johnmcollier: While it may be fine via the command line, most of the users will be interacting with CDQs via HAC, so that's where the latency really matters IMO.

Let's discuss when Elson's back.

johnmcollier (approved):

controllers/componentdetectionquery_controller_conditions.go
yangcao77: updated. removed the old `checkComponentName`, and use the `getComponentName`

maysunfaisal (approved):

maysunfaisal:
/hold

Wait until https://github.com/redhat-appstudio/infra-deployments/pull/662 is merged

PR #171, opened 2022-08-26
PLNSRVCE-624: Detect HACBS workflow

Michkov:
* Detect HACBS by checking if hacbs configmap exists
* When HACBS is enabled then:
  - select hacbs tekton bundle
  - enable PipelinesAsCode

Note: HACBS detection by configmap is temporary solution, will be
changed to detection based on APIBinding API in KCP environment.

### DIFF
--- a/gitops/prepare/prepare.go
+++ b/gitops/prepare/prepare.go
@@ -31,12 +31,18 @@ const (
 	BuildBundleConfigMapName = "build-pipelines-defaults"
 	// data key within a configMap that holds the URL to a build bundle
 	BuildBundleConfigMapKey = "default_build_bundle"
+	HACBSBundleConfigMapKey = "hacbs_build_bundle"
+
 	// fallback bundle that will be used in case the bundle resolution fails
 	FallbackBuildBundle = "quay.io/redhat-appstudio/build-templates-bundle:8201a567956ba6d2095d615ea2c0f6ab35f9ba5f"
 	// default secret for app studio registry
 	RegistrySecret = "redhat-appstudio-registry-pull-secret"
 	// Pipelines as Code global configuration secret name
 	PipelinesAsCodeSecretName = "pipelines-as-code-secret"
+	// ConfigMap name for detection hacbs workflow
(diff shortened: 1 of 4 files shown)


psturc:
lgtm - once we have this and a follow-up PR in build-service merged, it'd be nice to update e2e. 2 reasons:
1. to cover this functionality by tests
2. to simplify the test code - [the current implementation for switching from appstudio to hacbs bundle (and back)](https://github.com/redhat-appstudio/e2e-tests/blob/61d7ecabc5d7646a47a4bd6af25ff7b74deb80cd/tests/build/build.go#L306-L340) is just.. not nice

So we will need changes in e2e-tests and build-definitions repos. If there are no objections, I'll create a jira for that and link it as a follow-up to PLNSRVCE-624

mmorhun (approved):
LGTM

gitops/prepare/prepare.go
```diff
@@ -31,12 +31,16 @@ const (
...
 	BuildBundleConfigMapKey = "default_build_bundle"
+	HACBSBundleConfigMapKey = "hacbs_build_bundle"
+
 	// fallback bundle that will be used in case the bundle resolution fails
 	FallbackBuildBundle = "quay.io/redhat-appstudio/build-templates-bundle:8201a567956ba6d2095d615ea2c0f6ab35f9ba5f"
 	// default secret for app studio registry
 	RegistrySecret = "redhat-appstudio-registry-pull-secret"
 	// Pipelines as Code global configuration secret name
 	PipelinesAsCodeSecretName = "pipelines-as-code-secret"
+	// ConfigMap name for detection hacbs workflow
```
gabemontero, line 42: I would add your 
```
Note: HACBS detection by configmap is temporary solution, will be
changed to detection based on APIBinding API in KCP environment.
```
note from your PR description in the comment here as well.  In the future folks will not think to look at the PR associated with this change to determine if this is temporary or not.

gabemontero (changes requested):
just a minor ask for a code comment @Michkov

PR #183, opened 2022-09-07
[DEVHAS-143] always append random string at end of component name

yangcao77:
The PR removes the kube get request to check naming conflict in existing components. Instead, it always append random 4-char string ([0-9][a-z][A-Z]) at the end to avoid conflict.


johnmcollier (approved):
👏

PR #189, opened 2022-09-21
sample path fix

joejstuart:
### What does this PR do?:
fixes path to the kcp sample application

None


johnmcollier:
We've removed the KCP code from HAS, so closing...

PR #194, opened 2022-10-05
[DEVHAS-141] Event Logging function

yangcao77:
### What does this PR do?:

This PR created a common log function for resource change event. it audit logging the event with multiple key-value pairs contains information on the `appstudio-component`(which is HAS), `namespace`, `resource type`, `action` and `resource name`

https://issues.redhat.com/browse/DEVHAS-141

### PR acceptance criteria:

- [ ] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact

### DIFF
--- a/controllers/componentdetectionquery_controller.go
+++ b/controllers/componentdetectionquery_controller.go
@@ -111,9 +111,16 @@ func (r *ComponentDetectionQueryReconciler) Reconcile(ctx context.Context, req c
 		source := componentDetectionQuery.Spec.GitSource
 		var devfileBytes, dockerfileBytes []byte
 		var clonePath string
+		var componentPath string
 		devfilesMap := make(map[string][]byte)
 		devfilesURLMap := make(map[string]string)
 		dockerfileContextMap := make(map[string]string)
+
+		context := source.Context
+		contextDir := context
+		if context == "" {
+			contextDir = "./"
+		}
 
 		if source.DevfileURL == "" {
 			isMultiComponent := false
@@ -121,21 +128,19 @@ func (r *ComponentDetectionQueryReconciler) Reconcile(ctx context.Context, req c
 			isDevfilePresent := false
 			log.Info(fmt.Sprintf("Attempting to read a devfile from the URL %s... %v", source.URL, req.NamespacedName))
 			// check if the project is multi-component or single-component
-			log.Info(fmt.Sprintf("Look for devfile or dockerfile at the root dir... %v", req.NamespacedName))
-			var gitURL string
 			if gitToken == "" {
-				gitURL, err = util.ConvertGitHubURL(source.URL, source.Revision)
+				gitURL, err := util.ConvertGitHubURL(source.URL, source.Revision, context)
+				log.Info(fmt.Sprintf("Look for devfile or dockerfile at the URL %s... %v", gitURL, req.NamespacedName))
 				if err != nil {
 					log.Error(err, fmt.Sprintf("Unable to convert Github URL to raw format, exiting reconcile loop %v", req.NamespacedName))
 					r.SetCompleteConditionAndUpdateCR(ctx, req, &componentDetectionQuery, err)
 					return ctrl.Result{}, nil
 				}
-
 				devfileBytes, dockerfileBytes = devfile.DownloadDevfileAndDockerfile(gitURL)
 			} else {
 				// Use SPI to retrieve the devfile from the private repository
 				// TODO - maysunfaisal also search for Dockerfile
-				devfileBytes, err = spi.DownloadDevfileUsingSPI(r.SPIClient, ctx, componentDetectionQuery.Namespace, source.URL, "main", "")
+				devfileBytes, err = spi.DownloadDevfileUsingSPI(r.SPIClient, ctx, componentDetectionQuery.Namespace, source.URL, "main", context)
 				if err != nil {
 					log.Error(err, fmt.Sprintf("Unable to curl for any known devfile locations from %s %v", source.URL, req.NamespacedName))
 				}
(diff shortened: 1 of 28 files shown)


pkg/log/log.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/go-logr/logr"
+)
+
+type ResourceChangeType string
+
+const (
+	ResourceCreate   ResourceChangeType = "Create"
+	ResourceUpdate   ResourceChangeType = "Update"
+	ResourceDelete   ResourceChangeType = "Delete"
+	ResourceComplete ResourceChangeType = "Complete"
```
kim-tsao, line 30: Can we get a comment describing what "Complete" means?  I think it's only used in the context of CDQ but it would be good to be clear about this

controllers/componentdetectionquery_controller.go
```diff
@@ -266,7 +267,9 @@ func (r *ComponentDetectionQueryReconciler) Reconcile(ctx context.Context, req c
 		log.Info(fmt.Sprintf("Deleting finished ComponentDetectionQuery resource %v", req.NamespacedName))
 		if err = r.Delete(ctx, &componentDetectionQuery); err != nil {
 			// Delete failed. Log the error but don't bother modifying the resource's status
-			log.Error(err, fmt.Sprintf("Unable to delete the ComponentDetectionQuery resource %v", req.NamespacedName))
+			logutil.LogAPIResourceChangeEvent(log, componentDetectionQuery.Name, "ComponentDetectionQuery", logutil.ResourceDelete, err)
+		} else {
+			logutil.LogAPIResourceChangeEvent(log, componentDetectionQuery.Name, "ComponentDetectionQuery", logutil.ResourceDelete, nil)
 		}
```
kim-tsao, line 273: Can we also log events that call spi.DownloadDevfileUsingSPI since we are initiating the call to access an end users' private repo

pkg/log/log.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+func LogAPIResourceChangeEvent(log logr.Logger, resourceName string, resource string, resourceChangeType ResourceChangeType, err error) {
+	log = log.WithValues("audit", "true")
+
+	if resource == "" {
+		log.Error(nil, "resource passed to LogAPIResourceChangeEvent was empty")
+		return
+	}
+	log = log.WithValues("name", resourceName).WithValues("resource", resource).WithValues("action", resourceChangeType)
```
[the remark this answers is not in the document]
yangcao77: namespace is added here, beginning of every controller: https://github.com/redhat-appstudio/application-service/blob/3c370f5f7c6f0963ba44c1fd57554fb379f9f30e/controllers/application_controller.go#L65

controllers/componentdetectionquery_controller.go
yangcao77: we can, but I'm not sure if it's a resource change event we want to monitor.

pkg/log/log.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"fmt"
+
+	"github.com/go-logr/logr"
+)
+
+type ResourceChangeType string
+
+const (
+	ResourceCreate   ResourceChangeType = "Create"
```
johnmcollier, line 27: How is `Create` / `Update` / `Delete` being defined in the context of audit logging? Is it "Resource xyz was created" on Kubernetes, or "The Controller has finished reconciling resource xyz" for the first time

johnmcollier: If it's the former, we can define custom Create/Update/Delete funcs that does the event log before calling reconcile:

e.g.

https://github.com/redhat-appstudio/application-service/blob/3c370f5f7c6f0963ba44c1fd57554fb379f9f30e/controllers/applicationsnapshotenvironmentbinding_controller.go#L302-L312

pkg/log/log.go
```diff
@@ -0,0 +1,46 @@
...
+	ResourceUpdate   ResourceChangeType = "Update"
+	ResourceDelete   ResourceChangeType = "Delete"
```
johnmcollier, line 29: Audit logging for delete might be tricky unless we define a custom Delete event handler for the reconciler (see my other comment)

pkg/log/log.go
johnmcollier: CC @kim-tsao

controllers/componentdetectionquery_controller.go
kim-tsao: It's not a change event but we are calling a repo that we have privileged access too.  Maybe a new action called `ACCESS` would be appropriate?

pkg/log/log.go
kim-tsao: @johnmcollier  good point, we don't want to be logging in the reconcile loop.  I think it's reasonable to log based on the former

yangcao77: that's a good point! but how would we audit CDQ event then. when the cdq is created, it's not completed. and the cdq creation time is just start processing the detection.

pkg/log/log.go
yangcao77: yea. that's one thing I found tricky as except for CDQ, we don't have the delete action on other resource types.

pkg/log/log.go
maysunfaisal: So `ResourceChangeType` will probably be different from `ConditionChangeType`, whereby `ConditionChangeType` is a subset of `ResourceChangeType`

```
Conditions:
    Last Transition Time:  2022-10-11T17:23:44Z
    Message:               ComponentDetectionQuery is processing
    Reason:                Success
    Status:                True
    Type:                  Processing
    Last Transition Time:  2022-10-11T17:23:46Z
    Message:               ComponentDetectionQuery has successfully finished
    Reason:                OK
    Status:                True
    Type:                  Completed
```

maysunfaisal: `Processing` is not really a resource change but a condition change

yangcao77: I agree the `completed` is not a resource change. but since the common function accepts args which is defined as `ResourceChangeType`, we better not to complicated the situation just want to split the `ResourceChange` and `ConditionChange`, specifically for CDQ ONLY. We have discussed during the team meeting that  for CDQ, we will audit `completed` as a resource change event.

pkg/log/log.go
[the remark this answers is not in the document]
yangcao77: since using ctrl manager can only get the object `Namespace`, updated to namespace in the audit log.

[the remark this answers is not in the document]
johnmcollier: @yangcao77 we can also get the clustername with `logicalcluster.From(e.Object).String()` I believe

pkg/log/log.go
yangcao77: updated

pkg/log/log.go
yangcao77: updated

kim-tsao (approved):

PR #197, opened 2022-10-12
[DEVHAS 169] Always return devfile and dockerfile URLs on CDQ

maysunfaisal:

### What does this PR do?:

Always returns dockerfile and devfile URL in CDQ, so that HAC can use these CDQ return values for a Component create as the Build Service will need to know where the Devfile/Dockerfile URL is

[DEVHAS 169](https://issues.redhat.com/browse/DEVHAS-169)

### PR acceptance criteria:

- [x] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact

  

- Added tests to ensure we are checking for Devfile/Dockerfile URLs on CDQ return
- `oc get hcdq` to make sure the Devfile/Dockerfile URLs are always there


pkg/devfile/devfile.go
```diff
@@ -206,8 +206,8 @@ func CreateDevfileForDockerfileBuild(uri, context string) (data.DevfileData, err
 	return devfileData, nil
 }
 
-// DownloadDevfile downloads devfile from the various possible devfile locations in dir and returns the contents
-func DownloadDevfile(dir string) ([]byte, error) {
+// DownloadDevfile downloads devfile from the various possible devfile locations in dir and returns the contents and its context
+func DownloadDevfile(dir string) ([]byte, string, error) {
 	var devfileBytes []byte
 	var err error
 	validDevfileLocations := []string{Devfile, HiddenDevfile, HiddenDirDevfile, HiddenDirHiddenDevfile}
```
yangcao77, line 213: looks like the list has a priority. can we have test that a repo contains multiple valid devfiles, and should return the highest priority?

johnmcollier:
@maysunfaisal can you rebase?

pkg/devfile/devfile.go
maysunfaisal: updated all tests for this func

pkg/devfile/devfile.go
```diff
@@ -206,8 +206,8 @@ func CreateDevfileForDockerfileBuild(uri, context string) (data.DevfileData, err
 	return devfileData, nil
 }
 
-// DownloadDevfile downloads devfile from the various possible devfile locations in dir and returns the contents
-func DownloadDevfile(dir string) ([]byte, error) {
+// DownloadDevfile downloads devfile from the various possible devfile locations in dir and returns the contents and its context
+func DownloadDevfile(dir string) ([]byte, string, error) {
```
johnmcollier, line 210: `FindAndDownloadDevfile`?
`FindAndGetDevfile`?

johnmcollier (commented):
Just a couple questions

pkg/devfile/devfile.go
```diff
@@ -217,26 +217,27 @@ func DownloadDevfile(dir string) ([]byte, error) {
...
 
 // DownloadFile downloads the specified file
 func DownloadFile(file string) ([]byte, error) {
 	return util.CurlEndpoint(file)
 }
 
-// DownloadDevfileAndDockerfile attempts to download the  devfile and dockerfile from the root of the specified url
-func DownloadDevfileAndDockerfile(url string) ([]byte, []byte) {
+// DownloadDevfileAndDockerfile attempts to download and return the devfile, devfile context and dockerfile from the root of the specified url
+func DownloadDevfileAndDockerfile(url string) ([]byte, string, []byte) {
```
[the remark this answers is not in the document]
maysunfaisal: yeah they can be different, the context we pass in for a component. We assume devfile and dockerfile in this context but it could be somewhere else. This is when we do the detection for devfile and dockerfile for upto 1 nested directory. If we find both devfile and dockerfile, well and good. If we find only a dockerfile its a dockerfile component. If we find devfile but not a dockerfile we check if the dockerfile is defined in the devfile otherwise we use alizer to match a component from our devfile sample and use that dockerfile. Thats the gist of it.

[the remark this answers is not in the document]
maysunfaisal: So when this function is being called, it checks for the devfile at the context level. If we dont find it here, we walk the filepath later to see other directories.

johnmcollier (approved):

PR #198, opened 2022-10-13
[DEVHAS-178] Use MergePatch to update CDQ status

johnmcollier:
Signed-off-by: John Collier <jcollier@redhat.com>

### What does this PR do?:
~This PR makes sure we don't start reconciling CDQ resources on KCP until after the `kcp.dev/cluster` annotation has been added.~

Since KCP may attach a number of KCP-specific annotations and labels to the resource, this may cause conflicts when trying to update the resource:
```
2022-10-13T15:16:51.519Z    ERROR   controllers.ComponentDetectionQuery Unable to update ComponentDetectionQuery    {"ComponentDetectionQuery": "has-e2e-ndry/quarkus-component-e2e-madtwgalft", "clusterName": "test", "error": "Operation cannot be fulfilled on componentdetectionqueries.appstudio.redhat.com \"quarkus-component-e2e-madtwgalft\": the object has been modified; please apply your changes to the latest version and try again"}
```

Since we only allow CDQs to  run once by design, this may be problematic. Switching to a MergePatch when we update the CDQs status will prevent this from occurring.

https://issues.redhat.com/browse/DEVHAS-178

### DIFF
--- a/controllers/componentdetectionquery_controller.go
+++ b/controllers/componentdetectionquery_controller.go
@@ -86,11 +86,6 @@ func (r *ComponentDetectionQueryReconciler) Reconcile(ctx context.Context, req c
 		return ctrl.Result{}, err
 	}
 
-	// If on KCP, requeue if the "kcp.dev/cluster" annotation has not been added yet
-	if req.ClusterName != "" && componentDetectionQuery.GetAnnotations()["kcp.dev/cluster"] == "" {
-		return ctrl.Result{Requeue: true}, nil
-	}
-
 	// If there are no conditions attached to the CDQ, the resource was just created
 	if len(componentDetectionQuery.Status.Conditions) == 0 {
 		// Start the ComponentDetectionQuery, and update its status condition accordingly
--- a/controllers/componentdetectionquery_controller.go
+++ b/controllers/componentdetectionquery_controller.go
@@ -92,6 +92,9 @@ func (r *ComponentDetectionQueryReconciler) Reconcile(ctx context.Context, req c
 		log.Info(fmt.Sprintf("Checking to see if a component can be detected %v", req.NamespacedName))
 		r.SetDetectingConditionAndUpdateCR(ctx, req, &componentDetectionQuery)
 
+		// Create a copy of the CDQ, to use as the base when setting the CDQ status via mergepatch later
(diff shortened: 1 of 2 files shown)


controllers/componentdetectionquery_controller.go
```diff
@@ -86,6 +86,11 @@ func (r *ComponentDetectionQueryReconciler) Reconcile(ctx context.Context, req c
 		return ctrl.Result{}, err
 	}
 
+	// If on KCP, requeue if the "kcp.dev/cluster" annotation has not been added yet
+	if req.ClusterName != "" && componentDetectionQuery.GetAnnotations()["kcp.dev/cluster"] == "" {
```
maysunfaisal, line 90: Wondering if there is a predicate that ignores changes to annotations? What happens if kcp decides to update the annotation label 🤔

johnmcollier: Unlikely to happen, as the label is specific to the KCP workspace the resource is on. Plus, if a reconcile _was_ triggered, then our deletion logic would trigger and the old CDQ would get deleted.

maysunfaisal (approved):

maysunfaisal (approved):

controllers/componentdetectionquery_controller_conditions.go
```diff
@@ -23,29 +23,34 @@ import (
...
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	appstudiov1alpha1 "github.com/redhat-appstudio/application-service/api/v1alpha1"
 )
 
 func (r *ComponentDetectionQueryReconciler) SetDetectingConditionAndUpdateCR(ctx context.Context, req ctrl.Request, componentDetectionQuery *appstudiov1alpha1.ComponentDetectionQuery) {
 	log := r.Log.WithValues("ComponentDetectionQuery", req.NamespacedName).WithValues("clusterName", req.ClusterName)
 
+	patch := client.MergeFrom(componentDetectionQuery.DeepCopy())
```
johnmcollier, line 34: Since we set the "Detecting" condition right at the start, it doesn't matter what we use as the base, so a separate "originalCDQ" param isn't needed here

maysunfaisal (approved):

PR #243, opened 2023-01-19
[DEVHAS-121] Gitops repo SLI

kim-tsao:
Signed-off-by: Kim Tsao <ktsao@redhat.com>

### What does this PR do?:

- Initial instrumentation of metrics for Gitops repo creation.  Other metrics will be added over time
- Corresponding changes to set up the ServiceMonitor will need to be made in the infra deploy repo

https://issues.redhat.com/browse/DEVHAS-121

### PR acceptance criteria:

- [ ] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact


config/default/kustomization.yaml
```diff
@@ -21,7 +21,7 @@ bases:
 # [CERTMANAGER] To enable cert-manager, uncomment all sections with 'CERTMANAGER'. 'WEBHOOK' components are required.
 #- ../certmanager # Comment for KCP
 # [PROMETHEUS] To enable prometheus monitor, uncomment all sections with 'PROMETHEUS'.
-#- ../prometheus
+- ../prometheus
```
johnmcollier, line 24: What happens if Prometheus isn't installed on the Kubernetes cluster when you do `make deploy`?

pkg/github/github.go
```diff
@@ -44,13 +54,23 @@ func GenerateNewRepositoryName(displayName, namespace, clusterName string) strin
 func GenerateNewRepository(client *github.Client, ctx context.Context, orgName string, repoName string, description string) (string, error) {
 	isPrivate := false
 	appStudioAppDataURL := "https://github.com/" + orgName + "/"
-
+	metrics.GitOpsRepoCreationTotalReqs.Inc()
 	r := &github.Repository{Name: &repoName, Private: &isPrivate, Description: &description}
-	_, _, err := client.Repositories.Create(ctx, orgName, r)
+	_, resp, err := client.Repositories.Create(ctx, orgName, r)
+
+	if 500 <= resp.StatusCode && resp.StatusCode <= 599 {
```
johnmcollier, line 61: What about certain non 5xx response codes? E.g. 429, 403?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,48 @@
...
+// limitations under the License.
+
+package metrics
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+var (
```
johnmcollier, line 23: This might be nitpicking too much, but do we need tests for these counters? Is that even straightforward to do?

pkg/github/github.go
```diff
@@ -25,11 +25,21 @@ import (
...
 
 const AppStudioAppDataOrg = "redhat-appstudio-appdata"
 
+// ServerError is used to identify gitops repo creation failures caused by server errors
+type ServerError struct {
+	err error
+}
+
+func (e *ServerError) Error() string {
+	return fmt.Sprintf("failed to create gitops repo %s", e.err)
```
johnmcollier, line 40: Nit: Can you update the `TestGenerateNewRepository` function to check the error message? Will make codecov happier 🙂

johnmcollier (commented):
Generally looks good, mostly nits here and there

config/default/kustomization.yaml
kim-tsao: It doesn't break but I should comment it out since it install a ServiceMonitor that doesn't get used.  We actually define the SM on the infra-deploy repo

pkg/github/github.go
kim-tsao: I just made a generalization that all 5xx errors are server side errors that users cannot fix on their own.  429 is related to rate limiting which we should have in place and 403 could be an unauthorized request (user problem that should not factor into the SLI).  

Let me know if there's a better way to distinguish between user and system errors.

pkg/metrics/metrics.go
kim-tsao: Yes, I can add some coverage to the existing unit tests

johnmcollier (approved):

PR #246, opened 2023-01-23
[DEVHAS-228] Override local Dockerfile URIs in Component devfile when Dockerfile is external

johnmcollier:
### What does this PR do?:
Updates the Component controller so that when a DockerfileURL is passed in, if the Component devfile contains local Dockerfile URIs, those URIs will be updated to use the fullly qualified DockerfileURL.

Additionally, the CDQ controller has now been updated to no longer return a DockerfileURL if a Dockerfile is found locally within the detected Component, as the Build process will take care of resolving the Dockerfile.

https://issues.redhat.com/browse/DEVHAS-228

### PR acceptance criteria:

- [x] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact


pkg/devfile/devfile.go
```diff
@@ -250,3 +253,30 @@ func DownloadDevfileAndDockerfile(url string) ([]byte, string, []byte) {
...
+	devfileComponents, err := devfile.GetComponents(common.DevfileOptions{ComponentOptions: common.ComponentOptions{
+		ComponentType: v1alpha2.ImageComponentType,
+	}})
+	if err != nil {
+		return nil, err
+	}
+
+	for _, comp := range devfileComponents {
+		if comp.Image != nil && comp.Image.Dockerfile != nil {
+			// If the DockerfileURI is local (i.e. not a fully qualified URL) overwrite it to be the given Dockerfile URL, otherwise ignore it
```
maysunfaisal, line 268: Why not? If Component Source has DockerfileURL, we probably want to overwrite it always?

controllers/component_controller.go
```diff
@@ -308,6 +308,7 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 				return ctrl.Result{}, err
 			}
 
+			// If the DockerfileURL is set in the Git source - update any local references to the Dockerfile in the devfile to use the fully qualified URL
```
maysunfaisal, line 311: probably want to place this somewhere close to the actual line

pkg/devfile/devfile.go
johnmcollier: My thinking was that we wouldn't want to overwrite it, but there be cases where users pass in DockerfileURL to overwrite a fully qualified URL in the devfile that isn't resolvable or maybe private. I'll update it

johnmcollier: Updated.

controllers/component_controller.go
johnmcollier: Updated

johnmcollier:
@maysunfaisal Review comments addressed

maysunfaisal (approved):

PR #250, opened 2023-01-31
[DEVHAS-223] Set revision in componentStub if set in CDQ's spec

johnmcollier:
Signed-off-by: John Collier <jcollier@redhat.com>

### What does this PR do?:
Updates the componentStub returned by the CDQ to include the Git revision if `GitSource.revision` was set in the CDQ's spec. 

https://issues.redhat.com/browse/DEVHAS-223

### PR acceptance criteria:
Tests pass and ComponentStub now includes the proper Git revision if set in the CDQ spec.

### DIFF
--- a/controllers/componentdetectionquery_controller.go
+++ b/controllers/componentdetectionquery_controller.go
@@ -173,7 +173,7 @@ func (r *ComponentDetectionQueryReconciler) Reconcile(ctx context.Context, req c
 					return ctrl.Result{}, nil
 				}
 
-				err = util.CloneRepo(clonePath, source.URL, gitToken)
+				err = util.CloneRepo(clonePath, source.URL, source.Revision, gitToken)
 				if err != nil {
 					log.Error(err, fmt.Sprintf("Unable to clone repo %s to path %s, exiting reconcile loop %v", source.URL, clonePath, req.NamespacedName))
 					r.SetCompleteConditionAndUpdateCR(ctx, req, &componentDetectionQuery, copiedCDQ, err)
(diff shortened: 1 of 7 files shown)


yangcao77:
Also need to update https://github.com/redhat-appstudio/application-service/blob/ab24f58d9d38ae0e59e69b5b7291ab7a80e1ab79/controllers/componentdetectionquery_controller.go#L148

controllers/componentdetectionquery_controller.go
```diff
@@ -145,7 +145,7 @@ func (r *ComponentDetectionQueryReconciler) Reconcile(ctx context.Context, req c
 			} else {
 				// Use SPI to retrieve the devfile from the private repository
 				// TODO - maysunfaisal also search for Dockerfile
-				devfileBytes, err = spi.DownloadDevfileUsingSPI(r.SPIClient, ctx, componentDetectionQuery.Namespace, source.URL, "main", context)
+				devfileBytes, err = spi.DownloadDevfileUsingSPI(r.SPIClient, ctx, componentDetectionQuery.Namespace, source.URL, source.Revision, context)
```
yangcao77, line 148: I'm wondering if no `Revision` has been passed in, will this cause any error?
we might want to add more unit tests on `Revision =""`,  and `Revision=<other than main>`

johnmcollier: Yes good find. The problem is that the actual SPI client requires a running Kube cluster (with SPI installed) in order to function properly, so it's not something that we can really test in the controller tests.

johnmcollier: Updated.

yangcao77 (approved):
changes look good to me

PR #252, opened 2023-01-31
Re-attempt GitOps regeneration on err situation

maysunfaisal:

### What does this PR do?:

This PR adds the logic to the Component controller to re-attempt GitOps regeneration when GitOps generation errors out during create/update controller reconciles.

When GitOps generation errors out during a Component create/update reconcile, the Controller queues up again for reconciliation but our reconcile logic is not granular enough to check for all the steps. We check only if a devfile has been updated. Since in this case if a devfile is not updated, the GitOps regen will not be executed leaving the GitOps resources in the repository in a limbo unless the Component.Spec has been updated.

So during reconcile, we look at the GitOps conditions to make sure it has been successful in the prior reconcile attempts. If not, we attempt regeneration and return accordingly.

https://issues.redhat.com/browse/DEVHAS-211

### PR acceptance criteria:

- [ ] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact

  

TBD


maysunfaisal:
@johnmcollier note that I have also moved the line for updating the component devfile status before trying gitops generation. I confirmed this switch is working well ie; if gitops generation fails we will still have he latest devfile in component status

controllers/component_controller.go
```diff
@@ -434,8 +474,10 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 				return ctrl.Result{}, err
 			}
 
-			// Generate and push the gitops resources, if necessary.
 			component.Status.ContainerImage = component.Spec.ContainerImage
+			component.Status.Devfile = string(yamlHASCompData)
```
johnmcollier, line 478: Can you call `r.Status().Update()` after here?

johnmcollier (approved):
One small request

controllers/component_controller.go
maysunfaisal: Do we need to? Because I checked and made sure that our `func SetGitOpsGeneratedConditionAndUpdateCR` calls `err := r.Client.Status().Update(ctx, component)` irrespective of an err or not 🤔

maysunfaisal: It seems like the next steps are either generateGitops or SetUpdateConditionAndUpdateCR which both call the `r.Client.Status().Update()`

johnmcollier: It’s for the gitops job, it needs to Kubectl get the component since we don’t have access to it directly

johnmcollier: If the tests don’t pass with the change - don’t worry.

maysunfaisal: Understood, updated the PR!

maysunfaisal:
This PR should unblock https://github.com/redhat-appstudio/infra-deployments/pull/1251

johnmcollier (approved):

PR #253, opened 2023-01-31
disable kcp tests

yangcao77:
Signed-off-by: Stephanie <yangcao@redhat.com>

### What does this PR do?:

This PR remove kcp tooling installation. 
The kcp test has been disabled in github Actions. 

### PR acceptance criteria:

- [ ] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact


.github/workflows/kcp-test.yml
```diff
@@ -1,3 +1,4 @@
+# Currently disabled in github actions
```
johnmcollier, line 1: Would prefer we remove the github actions file altogether

yangcao77: I'm thinking if we are going back to support kcp in the future, we can still re-use this file. 
I have disable the github action already, so so leaving this file would not affect the ci.

johnmcollier: Yes, but can always go back in the Git history if we need to find it. Also, with how much churn occurs on the KCP API, it's likely this test won't be valid when/if we revisit KCP. I think I'd rather we remove it altogether to avoid any confusion.

johnmcollier (approved):

PR #260, opened 2023-02-07
[DEVHAS-245] Prevent a panic from occurring when SelectDevfileFromTypes returns an error

johnmcollier:
### What does this PR do?:
When `recognizer.SelectDevFileFromTypes` returns an error, it also returns `-1` for its index return value, which is not a valid index. Since we not checking the error and just directly returning the index within the DevfileTypes array and error, this could lead to a panic when index == -1. 

This PR simply adds an err check and returns appropriately before we attempt to access the index within the DevfileTypes array.

https://issues.redhat.com/browse/DEVHAS-245

### PR acceptance criteria:

- [x] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact

### DIFF
--- a/pkg/devfile/detect.go
+++ b/pkg/devfile/detect.go
@@ -258,6 +258,9 @@ func (a AlizerClient) Analyze(path string) ([]model.Language, error) {
 // SelectDevFileFromTypes is a wrapper call to Alizer's SelectDevFileFromTypes()
 func (a AlizerClient) SelectDevFileFromTypes(path string, devFileTypes []model.DevFileType) (model.DevFileType, error) {
 	index, err := recognizer.SelectDevFileFromTypes(path, devFileTypes)
+	if err != nil {
+		return model.DevFileType{}, err
+	}
 	return devFileTypes[index], err
 }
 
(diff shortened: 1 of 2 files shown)


pkg/devfile/detect.go
```diff
@@ -258,6 +258,9 @@ func (a AlizerClient) Analyze(path string) ([]model.Language, error) {
 // SelectDevFileFromTypes is a wrapper call to Alizer's SelectDevFileFromTypes()
 func (a AlizerClient) SelectDevFileFromTypes(path string, devFileTypes []model.DevFileType) (model.DevFileType, error) {
 	index, err := recognizer.SelectDevFileFromTypes(path, devFileTypes)
+	if err != nil {
+		return model.DevFileType{}, err
+	}
```
kim-tsao, line 263: Do we have test coverage for the case where a devfile is detected but there is an error? line 264

johnmcollier: Alizer only ever returns `-1` when there's an error

kim-tsao: ok, so you're saying the only time there's ever an error is when a -1 index is returned, otherwise it's nil - just checked the selectDevFileFromTypes implementation

kim-tsao (approved):

pkg/devfile/detect.go
johnmcollier: Yup. Your initial comment made me double check too 🙂

PR #266, opened 2023-02-10
Use Patch() to update SEB status

johnmcollier:
Changes the logic to update the SEB status from `Status().Update()` to `Status().Patch()`

### DIFF
--- a/controllers/applicationsnapshotenvironmentbinding_controller_conditions.go
+++ b/controllers/applicationsnapshotenvironmentbinding_controller_conditions.go
@@ -27,9 +27,16 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func (r *SnapshotEnvironmentBindingReconciler) SetConditionAndUpdateCR(ctx context.Context, req ctrl.Request, appSnapshotEnvBinding *appstudiov1alpha1.SnapshotEnvironmentBinding, patch client.Patch, createError error) {
+func (r *SnapshotEnvironmentBindingReconciler) SetConditionAndUpdateCR(ctx context.Context, req ctrl.Request, appSnapshotEnvBinding *appstudiov1alpha1.SnapshotEnvironmentBinding, createError error) {
 	log := r.Log.WithValues("SnapshotEnvironmentBinding", req.NamespacedName).WithValues("clusterName", req.ClusterName)
 
+	var currentSEB appstudiov1alpha1.SnapshotEnvironmentBinding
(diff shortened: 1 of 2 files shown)


controllers/applicationsnapshotenvironmentbinding_controller_conditions.go
```diff
@@ -24,30 +24,43 @@ import (
...
 			Reason:  "GenerateError",
 			Message: fmt.Sprintf("GitOps repository sync failed: %v", createError),
-		})
-		logutil.LogAPIResourceChangeEvent(log, appSnapshotEnvBinding.Name, "SnapshotEnvironmentBinding", logutil.ResourceCreate, createError)
+		}
+
 	}
+	meta.SetStatusCondition(&currentSEB.Status.GitOpsRepoConditions, condition)
+	logutil.LogAPIResourceChangeEvent(log, currentSEB.Name, "SnapshotEnvironmentBinding", logutil.ResourceCreate, createError)
+	currentSEB.Status.Components = appSnapshotEnvBinding.Status.Components
```
maysunfaisal, line 59: why do we this?
does it get overwritten?

maysunfaisal: we are doing `meta.SetStatusCondition(&currentSEB.Status.GitOpsRepoConditions, condition)`

johnmcollier: Because we're setting `Status.Components` earlier in the controller on the `appSnapshotEnvBinding` object, but it hasn't been applied server-side, so when we get the "current" object from Kube, it won't have it set. So since setting the status condition on the "current SEB" object, and updating it, I'm also making sure we copy over the `status.Components` status from earlier as well so that it gets applied server-side.

maysunfaisal: nevermind, its `currentSEB.Status.GitOpsRepoConditions` vs `currentSEB.Status.Components`. Sorry Friday afternoon brain fog.

So i guess its because we want to ensure we still have the `appSnapshotEnvBinding.Status.Components` changes we did before calling this function?

maysunfaisal (approved):

controllers/applicationsnapshotenvironmentbinding_controller_conditions.go
johnmcollier: Yup

flacatus:
@johnmcollier can u try to reopen a new PR?

johnmcollier:
😢

PR #273, opened 2023-02-22
[DEVHAS-264] Don't immediately display Application not ready error

johnmcollier:
### What does this PR do?:
This PR updates the logic in the Component controller for handling scenarios where the Application isn't reconciled before the Component is.

Previously, if a Component's associated Application resource was not finished reconciling, or had an error state associated with it, we would return an error message in the Component. However, in scenarios where the the Component and Application are created together, it's not ideal to immediately display that error message.

With this PR, when the Component controller encounters that error, we instead add a counter annotation to the Component. If the error occurs 5 consecutive times, we then return the original error in the Component status. Otherwise, if the error goes away, we don't display it.

https://issues.redhat.com/browse/DEVHAS-264

### PR acceptance criteria:

- [ ] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact

### DIFF
--- a/controllers/component_controller.go
+++ b/controllers/component_controller.go
@@ -74,6 +74,11 @@ type ComponentReconciler struct {
 	GitHubClient    *gh.Client
 }
 
+const (
+	applicationFailCounterAnnotation = "applicationFailCounter"
+	maxApplicationFailCount          = 5
+)
+
 //+kubebuilder:rbac:groups=appstudio.redhat.com,resources=components,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=appstudio.redhat.com,resources=components/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=appstudio.redhat.com,resources=components/finalizers,verbs=update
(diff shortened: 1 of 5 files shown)


controllers/component_controller.go
```diff
@@ -649,3 +653,22 @@ func (r *ComponentReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	}).
 		Complete(r)
 }
+
+func (r *ComponentReconciler) incrementCounterAndRequeue(log logr.Logger, ctx context.Context, req ctrl.Request, component *appstudiov1alpha1.Component, componentErr error) (ctrl.Result, error) {
+	if component.GetAnnotations() == nil {
+		component.ObjectMeta.Annotations = make(map[string]string)
+	}
+	count, err := getCounterAnnotation(applicationFailCounterAnnotation, component)
+	if count > 2 || err != nil {
```
yangcao77, line 662: so we allow up to 3 times retry for component creation due to application not ready failure. and application has 5 retrys?

better put some comment for easy understanding. also, do we want to document those details in book of stonesoup?

controllers/component_controller_test.go
```diff
@@ -373,7 +377,7 @@ var _ = Describe("Component controller", func() {
 			Eventually(func() bool {
 				k8sClient.Get(context.Background(), hasCompLookupKey, createdHasComp)
 				return len(createdHasComp.Status.Conditions) > 0 && createdHasComp.Status.Conditions[0].Reason == "OK"
-			}, timeout, interval).Should(BeTrue())
+			}, 4*timeout, interval).Should(BeTrue())
```
yangcao77, line 380: https://github.com/redhat-appstudio/application-service/blob/3c95c979794ddf571c3ece66783a168fca96873b/controllers/application_controller_test.go#L35

controllers/component_controller_test.go
```diff
@@ -359,12 +359,16 @@ var _ = Describe("Component controller", func() {
 			Eventually(func() bool {
 				k8sClient.Get(context.Background(), hasCompLookupKey, createdHasComp)
 				return len(createdHasComp.Status.Conditions) > 0
-			}, timeout, interval).Should(BeTrue())
+			}, 5*timeout, interval).Should(BeTrue())
```
yangcao77, line 362: any reason this one uses `5*timeout`? and the other one uses `4*timeout`?

controllers/component_controller.go
johnmcollier: The difference is because of the delayed requeues. Will add a comment to clarify

controllers/component_controller_test.go
johnmcollier: Nope, no reason in particular. Will update this

controllers/component_controller_test.go
johnmcollier: Good eye

johnmcollier: Updated

controllers/component_controller_test.go
johnmcollier: Updated.

controllers/component_controller.go
johnmcollier: Updated. 

Don't think these more fine-grained details need to go into the book of stonesoup, we're just slightly delaying when we consider the Component CR to have errored out in this scenario

yangcao77 (approved):
changes look good to me

PR #279, opened 2023-03-02
[DEVHAS-263] gitops generator check deploy cmd

yangcao77:
### What does this PR do?:

This PR updates the behavior when generating deployment-patch from devfile:

- If only one kubernetes component defined, directly use it. 
- if more than one kubernetes components defined, look for the one being referenced by default deploy command. 

https://issues.redhat.com/browse/DEVHAS-263

### PR acceptance criteria:

- [x] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact

### DIFF
--- a/pkg/devfile/devfile.go
+++ b/pkg/devfile/devfile.go
@@ -71,7 +71,6 @@ func GetResourceFromDevfile(log logr.Logger, devfileData data.DevfileData, deplo
 	if err != nil {
 		return parser.KubernetesResources{}, err
 	}
-	var component v1alpha2.Component
 
 	var appendedResources parser.KubernetesResources
 	k8sLabels := generateK8sLabels(compName, appName)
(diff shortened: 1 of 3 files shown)


yangcao77:
blocked by https://github.com/redhat-appstudio/application-service/pull/278

pkg/devfile/devfile.go
```diff
@@ -76,7 +76,14 @@ func GetResourceFromDevfile(log logr.Logger, devfileData data.DevfileData, deplo
 	k8sLabels := generateK8sLabels(compName, appName)
 	matchLabels := getMatchLabel(compName)
 
+	if len(kubernetesComponents) == 0 {
+		return parser.KubernetesResources{}, fmt.Errorf("the devfile has no kubernetes components defined, missing outerloop definition")
```
johnmcollier, line 80: I can't remember what was discussed around this.

What happens when this error is returned? If no Kubernetes components are defined, I feel like we should fall back on gitops resource generation from within HAS? Rather than erroring out

yangcao77: it will error out as it's not a valid outerloop devfile. and this error won't be triggered for a normal end-to-end use case. as this error will be detected and reported by the CDQ (change in https://github.com/redhat-appstudio/application-service/pull/278).

johnmcollier (approved):

PR #280, opened 2023-03-07
[DEVHAS-274] Always return local Dockerfile URI if Dockerfile present in repository

johnmcollier:
### What does this PR do?:
Removes the code that would change the local dockerfile URI to be absolute for Dockerfile components. Since we no longer want to provide an absolute URI in this component, this can be removed, and we can fall back on returning a local URI.

https://issues.redhat.com/browse/DEVHAS-274

### PR acceptance criteria:

- [ ] Documentation 

   

- [ ] Client Impact

### DIFF
--- a/controllers/componentdetectionquery_controller.go
+++ b/controllers/componentdetectionquery_controller.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"fmt"
 	"path"
-	"strings"
 
 	"go.uber.org/zap/zapcore"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


kim-tsao (approved):
Looks good, but we should probably get a couple of issues opened to update the API doc and add webhook validation to ensure URI is local

johnmcollier:
@kim-tsao The API will be updated in  https://github.com/redhat-appstudio/application-api/pull/36 - The URI can be either absolute or local, so no additional webhook validation should be needed.

PR #285, opened 2023-03-21
[DEVHAS-290] Support multiple tokens & Initialize go-github during reconcile

johnmcollier:
### What does this PR do?:
- Allow HAS to support using an arbitrary number of GitHub tokens for its Git operations.
    - A new environment variable, `GITHUB_TOKEN_LIST`, has been created for this. It can be set to a comma separated list of GitHub tokens. E.g. `GITHUB_TOKEN_LIST=some_token1,some_token2,some_token3`
    - `GITHUB_AUTH_TOKEN` remains in place in HAS, but I've removed its mention from the readme
- Moves Go-GitHub client initialization from `main.go` to each controller's `Reconcile()` function. When a client is initialized, a GitHub token will be randomly chosen, from the ones available to HAS.
    - the `GitHubToken` interface (which both `GitHubTokenClient` and `MockGitHubTokenClient` implement) provides a function for handling this.
    - Each controller's reconciler struct will have access to this

[](https://issues.redhat.com/browse/DEVHAS-290)

### PR acceptance criteria:

- [x] Unit/Functional tests

  

- [x] Documentation 

   

- [ ] Client Impact

### DIFF
--- a/controllers/application_controller.go
+++ b/controllers/application_controller.go
@@ -24,7 +24,6 @@ import (
 	"github.com/go-logr/logr"
 	"go.uber.org/zap/zapcore"
 
-	gh "github.com/google/go-github/v41/github"
 	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -48,10 +47,10 @@ import (
 // ApplicationReconciler reconciles a Application object
 type ApplicationReconciler struct {
 	client.Client
-	Scheme       *runtime.Scheme
-	Log          logr.Logger
-	GitHubClient *gh.Client
-	GitHubOrg    string
+	Scheme            *runtime.Scheme
+	Log               logr.Logger
+	GitHubTokenClient github.GitHubToken
+	GitHubOrg         string
 }
 
 //+kubebuilder:rbac:groups=appstudio.redhat.com,resources=applications,verbs=get;list;watch;create;update;patch;delete
@@ -89,6 +88,11 @@ func (r *ApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return reconcile.Result{}, err
 	}
 
+	ghClient, err := r.GitHubTokenClient.GetNewGitHubClient()
+	if err != nil {
+		log.Error(err, "Unable to create Go-GitHub client due to error")
+	}
+
 	// Check if the Application CR is under deletion
 	// If so: Remove the GitOps repo (if generated) and remove the finalizer.
 	if application.ObjectMeta.DeletionTimestamp.IsZero() {
@@ -100,7 +104,7 @@ func (r *ApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	} else {
 		if containsString(application.GetFinalizers(), appFinalizerName) {
 			// A finalizer is present for the Application CR, so make sure we do the necessary cleanup steps
-			if err := r.Finalize(&application); err != nil {
+			if err := r.Finalize(&application, ghClient); err != nil {
 				finalizeCounter, err := getCounterAnnotation(finalizeCount, &application)
 				if err == nil && finalizeCounter < 5 {
 					// The Finalize function failed, so increment the finalize count and return
@@ -139,7 +143,7 @@ func (r *ApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 			repoName := github.GenerateNewRepositoryName(application.Name, uniqueHash)
 
 			// Generate the git repo in the redhat-appstudio-appdata org
-			repoUrl, err := github.GenerateNewRepository(r.GitHubClient, ctx, r.GitHubOrg, repoName, "GitOps Repository")
+			repoUrl, err := github.GenerateNewRepository(ghClient, ctx, r.GitHubOrg, repoName, "GitOps Repository")
 			if err != nil {
 				log.Error(err, fmt.Sprintf("Unable to create repository %v", repoUrl))
 				r.SetCreateConditionAndUpdateCR(ctx, req, &application, err)
--- a/controllers/application_controller.go
+++ b/controllers/application_controller.go
@@ -88,7 +88,7 @@ func (r *ApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return reconcile.Result{}, err
 	}
 
-	ghClient, err := r.GitHubTokenClient.GetNewGitHubClient()
+	ghClient, _, err := r.GitHubTokenClient.GetNewGitHubClient()
 	if err != nil {
 		log.Error(err, "Unable to create Go-GitHub client due to error")
 		return reconcile.Result{}, err
--- a/controllers/application_controller.go
+++ b/controllers/application_controller.go
@@ -88,7 +88,7 @@ func (r *ApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return reconcile.Result{}, err
 	}
 
-	ghClient, _, err := r.GitHubTokenClient.GetNewGitHubClient()
+	ghClient, err := r.GitHubTokenClient.GetNewGitHubClient()
 	if err != nil {
 		log.Error(err, "Unable to create Go-GitHub client due to error")
 		return reconcile.Result{}, err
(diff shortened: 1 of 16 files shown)


controllers/applicationsnapshotenvironmentbinding_controller.go
```diff
@@ -99,6 +98,11 @@ func (r *SnapshotEnvironmentBindingReconciler) Reconcile(ctx context.Context, re
 
 	log.Info(fmt.Sprintf("Starting reconcile loop for %v %v", appSnapshotEnvBinding.Name, req.NamespacedName))
 
+	ghClient, err := r.GitHubTokenClient.GetNewGitHubClient()
+	if err != nil {
+		log.Error(err, "Unable to create Go-GitHub client due to error")
```
kim-tsao, line 103: ghclient is nil at this point. We should return and stop processing

controllers/component_controller.go
```diff
@@ -132,6 +132,11 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 
 	setCounterAnnotation(applicationFailCounterAnnotation, &component, 0)
 
+	ghClient, err := r.GitHubTokenClient.GetNewGitHubClient()
+	if err != nil {
+		log.Error(err, "Unable to create Go-GitHub client due to error")
```
kim-tsao, line 137: ghclient is nil at this point. We should return and stop processing

controllers/componentdetectionquery_controller.go
```diff
@@ -129,11 +128,17 @@ func (r *ComponentDetectionQueryReconciler) Reconcile(ctx context.Context, req c
 			context = "./"
 		}
 		if source.Revision == "" {
+			// Create a Go-GitHub client for checking the default branch
+			ghClient, err := r.GitHubTokenClient.GetNewGitHubClient()
+			if err != nil {
+				log.Error(err, "Unable to create Go-GitHub client due to error")
```
kim-tsao, line 134: ghclient is nil at this point. We should return and stop processing

pkg/github/token.go
```diff
@@ -0,0 +1,91 @@
...
+type GitHubTokenClient struct {
+}
+
+var Tokens []string
+
+// ParseGitHubTokens parses all of the possible GitHub tokens available to HAS and makes them available within the "github" package
+// This function should *only* be called once: at operator startup.
+func ParseGitHubTokens() error {
+	githubToken := os.Getenv("GITHUB_AUTH_TOKEN")
+	githubTokenList := os.Getenv("GITHUB_TOKEN_LIST")
```
kim-tsao, line 44: Is there a way we can make this a key value pair list?  It's easier to have an identifier associated with the token in order to log metrics about its usage

pkg/github/token_mock.go
```diff
@@ -0,0 +1,29 @@
...
+package github
+
+import (
+	"github.com/google/go-github/v41/github"
+)
+
+type MockGitHubTokenClient struct {
+}
+
+// GetNewGitHubClient intializes a new Go-GitHub client from a randomly selecte GitHub token available to HAS
```
kim-tsao, line 25: typos on selecte and retrieveing

johnmcollier: Fixed.

pkg/github/token.go
```diff
@@ -0,0 +1,91 @@
...
+		splitTokenList := strings.Split(githubTokenList, ",")
+		tokenList = append(tokenList, splitTokenList...)
+	}
+
+	Tokens = tokenList
+
+	return nil
+}
+
+// getRandomToken randomly retrieves
```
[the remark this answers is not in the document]
johnmcollier: Fixed.

pkg/github/token.go
```diff
@@ -0,0 +1,91 @@
...
+		return "", fmt.Errorf("no GitHub tokens initialized")
+	}
+	if len(Tokens) == 1 {
+		return Tokens[0], nil
+	}
+	return Tokens[rand.Intn(len(Tokens)-1)], nil
+}
+
+// GetNewGitHubClient intializes a new Go-GitHub client from a randomly selecte GitHub token available to HAS
+// If an error is encountered retrieveing the token, or initializing the client, an error is returned
```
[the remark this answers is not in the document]
johnmcollier: Oops, good catch. Fixed.

pkg/github/token.go
johnmcollier: Updated as discussed on today's call.

`Tokens` will be a map where the key points to the token name, and the value is the token itself. For tokens set through the legacy GITHUB_AUTH_TOKEN environment variable (which does not used named tokens), the key `GITHUB_AUTH_TOKEN` is used.

controllers/component_controller.go
johnmcollier: Fixed.

controllers/applicationsnapshotenvironmentbinding_controller.go
johnmcollier: Fixed

controllers/application_controller.go
```diff
@@ -89,6 +88,11 @@ func (r *ApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return reconcile.Result{}, err
 	}
 
+	ghClient, err := r.GitHubTokenClient.GetNewGitHubClient()
+	if err != nil {
+		log.Error(err, "Unable to create Go-GitHub client due to error")
```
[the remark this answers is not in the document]
johnmcollier: Oops, there should've been a return statement there. Fixed.

pkg/github/token.go
```diff
@@ -0,0 +1,123 @@
...
+// getRandomToken randomly retrieves a token from all of the tokens available to HAS
+func getRandomToken() (string, error) {
+	if len(Tokens) == 0 {
+		return "", fmt.Errorf("no GitHub tokens initialized")
+	}
+	var index int
+	if len(Tokens) == 1 {
+		index = 0
+	} else {
+		index = rand.Intn(len(Tokens) - 1)
```
kim-tsao, line 92: I don't think we need to subtract 1 since [rand.Intn](https://pkg.go.dev/math/rand#Intn) takes care of that by selecting a number between [0, n).  If `Tokens` is length 2, then 0 and 1 would be returned.  If we subtract 1 then 0 i.e. the first token would always be returned.   Having said that, I guess we don't need to check for length 1 and just have the statement `index=rand.Intn(len(Tokens))` take care of it all

pkg/github/token.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/gofri/go-github-ratelimit/github_ratelimit"
+	"github.com/google/go-github/v41/github"
+	"golang.org/x/oauth2"
+)
+
+type GitHubToken interface {
+	GetNewGitHubClient() (*github.Client, error)
+}
+
+type GitHubTokenClient struct {
```
kim-tsao, line 35: Can we add a field here for the token name?  I need to be able to access it in the controllers when I add the metrics

pkg/github/token.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+		i++
+	}
+	return token, nil
+}
+
+// GetNewGitHubClient intializes a new Go-GitHub client from a randomly selected GitHub token available to HAS
+// If an error is encountered retrieving the token, or initializing the client, an error is returned
+func (g GitHubTokenClient) GetNewGitHubClient() (*github.Client, error) {
+	ghToken, err := getRandomToken()
```
kim-tsao, line 109: I think we will need the token name returned here too so it can be set in the `GitHubTokenClient`

pkg/github/token.go
johnmcollier: I added it in an earlier commit here: https://github.com/redhat-appstudio/application-service/blob/7c07b8d5803ae91781b7155378a64cac36036404/README.md#creating-a-github-secret-for-has

pkg/github/token.go
kim-tsao: that can work too

pkg/github/token.go
johnmcollier: Yes, that's a lot better. Dunno why I didn't think of that haha.

Updated.

pkg/github/token.go
johnmcollier: Updated.

kim-tsao (approved):
Looks good, thanks for pulling together a solution so quickly

pkg/github/token.go
johnmcollier: We can't add it to GitHubTokenClient, as it's just used for retrieving the token and generating the Go-GitHub clients

But I added a wrapper struct to the Go-GitHub client, which now includes the token name, along with the client itself.

kim-tsao: ok, that sounds good

kim-tsao (approved):

PR #289, opened 2023-03-29
[DEVHAS-223 Redux] Ensure revision is set for Dockerfile-only components

johnmcollier:
### What does this PR do?:
In https://github.com/redhat-appstudio/application-service/pull/250, we addressed the bug where the source revision for Components was not being returned in the CDQ status. However, it seems that we were only setting the revision for Components with a Devfile present. For Dockerfile-only components, the revision was not being set, as the revision needed to be set in a different part of code.

This PR updates the CDQ logic so that the revision is also returned for Dockerfile components, and updates test cases to ensure this scenario is covered.

https://issues.redhat.com/browse/DEVHAS-223

### PR acceptance criteria:

- [x] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact

### DIFF
--- a/controllers/componentdetectionquery_controller_test.go
+++ b/controllers/componentdetectionquery_controller_test.go
@@ -989,7 +989,8 @@ var _ = Describe("Component Detection Query controller", func() {
 					},
 					Spec: appstudiov1alpha1.ComponentDetectionQuerySpec{
 						GitSource: appstudiov1alpha1.GitSource{
-							URL: "https://github.com/maysunfaisal/python-src-docker",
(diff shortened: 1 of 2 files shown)


johnmcollier:
Confirmed this works now:
```
      componentStub:
        application: insert-application-name
        componentName: devfile-sample-weom
        resources: {}
        source:
          git:
            context: ./
            dockerfileUrl: ./Dockerfile
            revision: no-devfile
            url: https://github.com/christianvogt/devfile-sample
```

kim-tsao (approved):

PR #304, opened 2023-04-24
[DEVHAS-310] fix: Requeue if status update fails for Components

johnmcollier:
### What does this PR do?:
When a Custom Resource we're watching fails to update its status, we're only logging the error and not requeueing, which means it's possible (albeit very unlikely) for a given resource to get stuck. Generally this hasn't been an issue, and we haven't seen issues get stuck, but we've recently seen it happen more frequently for Component resources, as a result of changes involving the [image controller](https://github.com/redhat-appstudio/image-controller). 
 
To resolve this, I've updated the functions in `component_controller_conditions.go` to do two things:
1) If the status update fails, we get the latest version of the resource from the cached API server and re-attempt the status update
    - This may not always be the most up to date version, so the following step is needed too
2) If the status update fails a second time, we return with an error. In most cases, we're already requeueing after encountering an error. So we only need to make sure we requeue in the cases that status update failed after a successful reconcile.

**Note:** Keeping in mind that https://issues.redhat.com/browse/DEVHAS-310 is a blocker, and only the Component is particularly affected now, I've scoped my change here just to the Component controller to keep the change small.

https://issues.redhat.com/browse/DEVHAS-310

### PR acceptance criteria:

- [x] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact

### DIFF
--- a/controllers/component_controller.go
+++ b/controllers/component_controller.go
@@ -60,6 +60,7 @@ import (
 // ComponentReconciler reconciles a Component object
 type ComponentReconciler struct {
 	client.Client
+	NonCachingClient  client.Client
 	Scheme            *runtime.Scheme
 	Log               logr.Logger
 	GitHubOrg         string
@@ -187,17 +188,20 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 			if err != nil {
 				errMsg := fmt.Sprintf("Unable to parse the devfile from Component status and re-attempt GitOps generation, exiting reconcile loop %v", req.NamespacedName)
 				log.Error(err, errMsg)
-				r.SetGitOpsGeneratedConditionAndUpdateCR(ctx, req, &component, fmt.Errorf("%v: %v", errMsg, err))
+				_ = r.SetGitOpsGeneratedConditionAndUpdateCR(ctx, req, &component, fmt.Errorf("%v: %v", errMsg, err))
 				return ctrl.Result{}, err
 			}
 			if err := r.generateGitops(ctx, ghClient, req, &component, compDevfileData); err != nil {
 				errMsg := fmt.Sprintf("Unable to generate gitops resources for component %v", req.NamespacedName)
 				log.Error(err, errMsg)
-				r.SetGitOpsGeneratedConditionAndUpdateCR(ctx, req, &component, fmt.Errorf("%v: %v", errMsg, err))
+				_ = r.SetGitOpsGeneratedConditionAndUpdateCR(ctx, req, &component, fmt.Errorf("%v: %v", errMsg, err))
 				return ctrl.Result{}, err
 			} else {
 				log.Info(fmt.Sprintf("GitOps re-generation successful for %s", component.Name))
-				r.SetGitOpsGeneratedConditionAndUpdateCR(ctx, req, &component, nil)
+				err := r.SetGitOpsGeneratedConditionAndUpdateCR(ctx, req, &component, nil)
+				if err != nil {
+					return ctrl.Result{}, err
+				}
 				isGitOpsRegenSuccessful = true
 			}
 		} else if condition.Type == "Updated" && condition.Reason == "Error" && condition.Status == metav1.ConditionFalse {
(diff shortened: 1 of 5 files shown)


flacatus (approved):

controllers/component_controller.go
```diff
@@ -60,6 +60,7 @@ import (
 // ComponentReconciler reconciles a Component object
 type ComponentReconciler struct {
 	client.Client
+	NonCachingClient  client.Client
```
maysunfaisal, line 63: is this only used in test?

controllers/component_controller.go
```diff
@@ -278,21 +288,24 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
 				}
 
 				devfileBytes, err = yaml.Marshal(devfileData)
 				if err != nil {
 					log.Error(err, fmt.Sprintf("Unable to marshal devfile, exiting reconcile loop %v", req.NamespacedName))
-					r.SetCreateConditionAndUpdateCR(ctx, req, &component, err)
+					err = r.SetCreateConditionAndUpdateCR(ctx, req, &component, err)
+					if err != nil {
+						return ctrl.Result{}, err
+					}
```
maysunfaisal, line 308: wondering why we are specifically checking for the condition err case here

controllers/component_controller_conditions.go
```diff
@@ -28,80 +28,151 @@ import (
...
+	meta.SetStatusCondition(&component.Status.Conditions, condition)
 
 	err := r.Client.Status().Update(ctx, component)
 	if err != nil {
-		log.Error(err, "Unable to update Component")
+		// Retry, and if still fails, then return an error
+		curComponent := &appstudiov1alpha1.Component{
+			TypeMeta: metav1.TypeMeta{
+				APIVersion: "appstudio.redhat.com/v1alpha1",
+				Kind:       "Application",
```
maysunfaisal, line 59:
```suggestion
			TypeMeta: metav1.TypeMeta{
				APIVersion: "appstudio.redhat.com/v1alpha1",
				Kind:       "Component",
```
should be? Or do we even need to set TypeMeta?

controllers/component_controller_conditions.go
```diff
@@ -28,80 +28,151 @@ import (
...
 
+	meta.SetStatusCondition(&component.Status.Conditions, condition)
 	err := r.Client.Status().Update(ctx, component)
 	if err != nil {
-		log.Error(err, "Unable to update Component")
+		// Retry, and if still fails, then return an error
+		curComponent := &appstudiov1alpha1.Component{
+			TypeMeta: metav1.TypeMeta{
+				APIVersion: "appstudio.redhat.com/v1alpha1",
+				Kind:       "Application",
```
maysunfaisal, line 110: same here

controllers/component_controller_conditions.go
johnmcollier: Lol oops.

Should be removed. The typemeta doesn't factor into what's retrieved. This was leftover from testing.

johnmcollier: Fixed

controllers/component_controller_conditions.go
johnmcollier: Fixed

controllers/component_controller_conditions.go
```diff
@@ -28,80 +28,151 @@ import (
...
+	meta.SetStatusCondition(&component.Status.Conditions, condition)
 
 	err := r.Client.Status().Update(ctx, component)
 	if err != nil {
-		log.Error(err, "Unable to update Component")
+		// Retry, and if still fails, then return an error
+		curComponent := &appstudiov1alpha1.Component{
+			TypeMeta: metav1.TypeMeta{
+				APIVersion: "appstudio.redhat.com/v1alpha1",
+				Kind:       "Application",
```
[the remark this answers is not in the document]
johnmcollier: Fixed

johnmcollier:
@maysunfaisal comments addressed

maysunfaisal (approved):

PR #307, opened 2023-04-28
feat: Consider primary and secondary rate limits when choosing tokens

johnmcollier:
### What does this PR do?:
This PR adds checks for Primary and Secondary rate limits when choosing tokens.

When HAS goes to initialize a new GitHubClient object, it now does the following:

For **Primary Rate Limits**:
   - HAS will now check the `Core` and `Search` limits for the user associated with the client
   - If the `Core` limit has less than 10 available requests (out of 5000), or if the `Search` limit has 0 or 1 available requests (out of 30), HAS will attempt to select another token
   - If there are no further tokens available that aren't rate limited, an error will be returned

For **Secondary Rate Limits**:
- A new field has been added to the `GitHubClient` struct, `SecondaryRateLimit`
- When [go-github-ratelimit](https://github.com/gofri/go-github-ratelimit/) detects a secondary rate limit, the `SecondaryRateLimit` rate limited field will be set to true
   - By way of a custom callback function
- When a token is selected, after checking the primary rate limit, HAS will check the `SecondaryRateLimit` field.
   - If it's true, HAS will attempt to retrieve another token until there are no more un-ratelimited tokens available
   - If it's false, the token is returned

Some other changes that were made to accommodate this:
- The list of tokens that we keep as been changed from `map[string]string` to map[string]*GitHubClient
   - The key, the token/client name remains the same. But changing the value to `*GitHubClient` allows us to easily retrieve the GitHubClient object associated with the client.
- `createGitHubClientFromToken` now takes in `http.RoundTripper` as an additional parameter. This allows us to configure `go-github-ratelimit` with the mock Go-GitHub client we use in tests

https://issues.redhat.com/browse/DEVHAS-344
https://issues.redhat.com/browse/DEVHAS-345

### PR acceptance criteria:

- [x] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact


pkg/github/token.go
```diff
@@ -131,5 +184,28 @@ func (g GitHubTokenClient) GetNewGitHubClient(token string) (GitHubClient, error
...
+	ghClient := Clients[ghClientName]
+
+	// Start a goroutine that marks the given client as rate limited and sleeps for 'TotalSleepTime'
+	go func(client *GitHubClient) {
+		client.SecondaryRateLimit.mu.Lock()
+		client.SecondaryRateLimit.isLimitReached = true
+		client.SecondaryRateLimit.mu.Unlock()
+
+		// Sleep until the rate limit is over
+		time.Sleep(*cbContext.TotalSleepTime)
```
gofri, line 205: I think you want to use cbContext.SleepUntil here, as TotalSleepTime accumulates the sleep durations in case of multiple hits of secondary rate limit.
So probably looking for:
`time.Sleep(time.Until(*cbContext.SleepUntil))` 

I'll add documentation for the different fields in go-github-ratelimit to avoid this kind of confusion.

johnmcollier: Hmm, interesting. In my tests, I'm finding that `time.Until(*cbContext.SleepUntil)` is a negative number, and the callback function is already sleeping before I call my `time.Sleep` 🤔

pkg/github/token.go
```diff
@@ -70,59 +78,104 @@ func ParseGitHubTokens() error {
...
 }
 
-// getRandomToken randomly retrieves a token from all of the tokens available to HAS
+// getRandomClient randomly retrieves a token from all of the tokens available to HAS
 // It returns the token and the name/key of the token
-func getRandomToken() (string, string, error) {
-	if len(Tokens) == 0 {
-		return "", "", fmt.Errorf("no GitHub tokens initialized")
+func getRandomClient(clientPool map[string]*GitHubClient) (*GitHubClient, error) {
+	if len(clientPool) == 0 {
```
kim-tsao, line 101: Will we ever get primary rate limited if we're doing preemptive checks?  If not, then I'll have to change the way the metrics work

kim-tsao: This will work for secondary rate limiting too.  Can we pass in a couple of booleans to indicate if we've exhausted the tokens for primary or secondary rate limiting?

pkg/github/token_test.go
```diff
@@ -132,46 +162,33 @@ func TestGetNewGitHubClient(t *testing.T) {
...
+			name:    "Mock client",
+			client:  MockGitHubTokenClient{},
+			wantErr: false,
 		},
 		{
 			name:          "Passed in token",
-			client:        ghTokenClient,
-			passedInToken: fakeToken,
+			client:        GitHubTokenClient{},
+			passedInToken: "fake-token", // Use an empty token here instead of a fake token string.
```
kim-tsao, line 172: Is `passedInToken` supposed to be an empty string or is the comment incorrect?

controllers/application_controller.go
```diff
@@ -229,22 +232,26 @@ func (r *ApplicationReconciler) SetupWithManager(ctx context.Context, mgr ctrl.M
...
-		},
-		DeleteFunc: func(e event.DeleteEvent) bool {
-			log := log.WithValues("namespace", e.Object.GetNamespace())
-			logutil.LogAPIResourceChangeEvent(log, e.Object.GetName(), "Application", logutil.ResourceDelete, nil)
-			return false
-		},
-	}).
+		For(&appstudiov1alpha1.Application{}).
+		WithOptions(controller.Options{
+			RateLimiter: workqueue.NewItemExponentialFailureRateLimiter(time.Duration(1*time.Second), time.Duration(1000*time.Second)),
```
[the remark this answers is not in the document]
johnmcollier: Component controller already has an alternate rate limiter set up, the CDQ controller doesn't need one (as it only reconciles resources once), and the current rate limiter works fine for SEBs

pkg/github/token.go
johnmcollier: No, the Rate Limit API endpoint is not rate limited.

> This will work for secondary rate limiting too. Can we pass in a couple of booleans to indicate if we've exhausted the tokens for primary or secondary rate limiting?

I'm not sure what you mean here? What would the booleans be used for?

controllers/component_controller.go
```diff
@@ -527,7 +527,7 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 
 // generateGitops retrieves the necessary information about a Component's gitops repository (URL, branch, context)
 // and attempts to use the GitOps package to generate gitops resources based on that component
-func (r *ComponentReconciler) generateGitops(ctx context.Context, ghClient github.GitHubClient, req ctrl.Request, component *appstudiov1alpha1.Component, compDevfileData data.DevfileData) error {
+func (r *ComponentReconciler) generateGitops(ctx context.Context, ghClient *github.GitHubClient, req ctrl.Request, component *appstudiov1alpha1.Component, compDevfileData data.DevfileData) error {
```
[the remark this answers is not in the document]
johnmcollier: Removed.

pkg/github/token.go
kim-tsao: That we exhausted the tokens for primary rate limit or secondary rate limits.  I can alternatively do a check on the length of the clientPool before the recursive call but seeing how there's a check done at the beginning, I thought we could avoid duplication.

johnmcollier: Right okay. So it's for metrics? What if we just add the necessary lines when we detect the primary/secondary rate limiting, instead of using bools to track?

kim-tsao: yeah, that's fine.  I can workaround it.  I was planning to add metrics if we are close to exhausting (~2 tokens left) and when we have no tokens left so we can get the alerts

johnmcollier:
Let's wait on https://github.com/google/go-github/pull/2775 before merging

/hold

johnmcollier:
After further discussion - we've decided to merge before https://github.com/google/go-github/pull/2775 is in, as it will be beneficial for us regardless.

/unhold

PR #308, opened 2023-05-01
[DEVHAS-320]remove hard coded main branch

yangcao77:
### What does this PR do?:

This PR updates the component controller and use revision set instead of hard-coded "main"

https://issues.redhat.com/browse/DEVHAS-320

### PR acceptance criteria:

- [ ] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact

### DIFF
--- a/controllers/component_controller.go
+++ b/controllers/component_controller.go
@@ -273,7 +273,7 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 					devfileLocation = gitURL + string(os.PathSeparator) + devfileLocation
 				} else {
 					// Use SPI to retrieve the devfile from the private repository
-					devfileBytes, err = spi.DownloadDevfileUsingSPI(r.SPIClient, ctx, component.Namespace, source.GitSource.URL, "main", context)
+					devfileBytes, err = spi.DownloadDevfileUsingSPI(r.SPIClient, ctx, component.Namespace, source.GitSource.URL, source.GitSource.Revision, context)
 					if err != nil {
 						log.Error(err, fmt.Sprintf("Unable to download from any known devfile locations from %s %v", source.GitSource.URL, req.NamespacedName))
 						_ = r.SetCreateConditionAndUpdateCR(ctx, req, &component, err)
--- a/controllers/component_controller.go
+++ b/controllers/component_controller.go
@@ -253,6 +253,33 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 				gitToken = string(gitSecret.Data["password"])
 			}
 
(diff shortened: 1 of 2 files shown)


johnmcollier:
I don't think this picked up my change from https://github.com/redhat-appstudio/infra-deployments/pull/1755 yet

/retest

controllers/component_controller.go
```diff
@@ -273,7 +273,7 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 					devfileLocation = gitURL + string(os.PathSeparator) + devfileLocation
 				} else {
 					// Use SPI to retrieve the devfile from the private repository
-					devfileBytes, err = spi.DownloadDevfileUsingSPI(r.SPIClient, ctx, component.Namespace, source.GitSource.URL, "main", context)
+					devfileBytes, err = spi.DownloadDevfileUsingSPI(r.SPIClient, ctx, component.Namespace, source.GitSource.URL, source.GitSource.Revision, context)
```
johnmcollier, line 276: When Components are created via HAC, this will work fine since the CDQ will always set the revision in the Component's git source, but outside of HAC, it may not be guaranteed that a CDQ preceeded Component creation, and I don't think we should assume that.

Can we attempt branch detection with the token from SPI if `source.GitSource.Revision` is unset?

johnmcollier (changes requested):

controllers/component_controller.go
yangcao77: for a production end-to-end, is that allowed to create any component outside of HAC? i.e. without running CDQ? I didn't add any check for the value or even default branch detection since that should be already done in CDQ, and I do not want to duplicate the process if a valid use case will be `application creation -> CDQ -> component creation`

johnmcollier: There's nothing outside of HAC that would stop the user creating Components without CDQs. E.g. in a future scenario where users have CLI access.

Since the Component spec doesn't say that a CDQ must be run before hand, I'd rather we not rely on that assumption. There's

yangcao77: updated

kim-tsao: Just to note, this will introduce additional go-github api calls which we wanted to avoid.   What will happen if we don't do branch detection?  While there's nothing stopping a user from making an  API call, it's still unsupported so if the call is made out of order and results in an error, can we issue a message instead? something like "make sure revision is specified"?

kim-tsao (approved):
As discussed in our meeting, branch detection will only occur if the component is created outside of the HAC UI, bypassing CDQ.  Since this represents a small percentage of use cases, it will not incur additional and unnecessary calls in the mainstream path.

PR #309, opened 2023-05-01
(Alternate) feat: Consider primary and secondary rate limits when choosing tokens

johnmcollier:
### What does this PR do?:
Alternate to https://github.com/redhat-appstudio/application-service/pull/307. Instead of using a callback function to handle the secondary rate limiting, we check the response from GitHub, and manually handle it there.

This PR adds checks for Primary and Secondary rate limits when choosing tokens.

When HAS goes to initialize a new GitHubClient object, it now does the following:

For **Primary Rate Limits**:
   - HAS will now check the `Core` and `Search` limits for the user associated with the client
   - If the `Core` limit has less than 10 available requests (out of 5000), or if the `Search` limit has 0 or 1 available requests (out of 30), HAS will attempt to select another token
   - If there are no further tokens available that aren't rate limited, an error will be returned

For **Secondary Rate Limits**:
- A new field has been added to the `GitHubClient` struct, `SecondaryRateLimit`
- When [go-github-ratelimit](https://github.com/gofri/go-github-ratelimit/) detects a secondary rate limit, the `SecondaryRateLimit` rate limited field will be set to true
   - By way of a custom callback function
- When a token is selected, after checking the primary rate limit, HAS will check the `SecondaryRateLimit` field.
   - If it's true, HAS will attempt to retrieve another token until there are no more un-ratelimited tokens available
   - If it's false, the token is returned

https://issues.redhat.com/browse/DEVHAS-344
https://issues.redhat.com/browse/DEVHAS-345

### PR acceptance criteria:

- [x] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact


johnmcollier:
This likely isn't needed anymore, but keeping this PR open as an alternate for the meanwhile

PR #312, opened 2023-05-03
[DEVHAS-350] update error check in port detection

yangcao77:
### What does this PR do?:

update error check in port detection, do not return error upon port detection failure

https://issues.redhat.com/browse/DEVHAS-350

### PR acceptance criteria:

- [ ] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact


pkg/devfile/detect.go
```diff
@@ -252,10 +252,9 @@ func AnalyzePath(a Alizer, localpath, context, devfileRegistryURL string, devfil
 	if !isDevfilePresent && isDockerfilePresent {
 		// Still invoke alizer to detect the ports from the component
 		_, _, _, detectedPorts, err := AnalyzeAndDetectDevfile(a, localpath, devfileRegistryURL)
-		if err != nil {
-			return err
+		if err == nil {
```
johnmcollier, line 255: Maybe add a log statement if the err is nil?

yangcao77:
> Looks good. Could we add a test case for this?

Since we updated to not return any error upon failure, it's not easy to using the mocked Alizer client to test with that. 
we have the issue now to switch to use actual Alizer client instead of the mocked one, https://issues.redhat.com/browse/DEVHAS-347. I'm thinking it will be better to add more CDQ test cases, especially for error cases, after that work has been done

pkg/devfile/detect.go
```diff
@@ -252,10 +252,11 @@ func AnalyzePath(a Alizer, localpath, context, devfileRegistryURL string, devfil
...
+			componentPortsMapFromRepo[context] = detectedPorts
+		} else {
+			log.Info("failed to detect port from context: %v, error: %v", context, err)
```
kim-tsao, line 258: should this be formatted?  `log.Info(fmt.Sprintf("failed to detect port from context: %v, error: %v", context, err))` or  alternatively follow the key/value pair logging convention for that API.

yangcao77: oops. updated

kim-tsao (approved):

PR #317, opened 2023-05-05
Don't allow secondary rate limit callback to panic

johnmcollier:
### What does this PR do?:
The secondary rate limit callback function we use can potentially panic if the Go-GitHub client name is not set in the context before the (rate limited) request is made. This PR takes the following steps to avoid that:
- The Go-GitHub client name is added to the context as soon as we initialize the Go-GitHub client
- nil pointer checks for when the Go-GitHub client is either unset or points to a non-existent client
   - There's nothing we can do when its nil, so logging and returning is preferable to panicking.
- Updated test cases to ensure that we don't panic in these scenarios


kim-tsao (approved):
/lgtm

PR #318, opened 2023-05-06
chore: Regenerate dashboard CM name

gbenhaim:
This will cause an update to the Dashboard CR that will get reconciled by the Grafana operator.


kim-tsao (approved):
/lgtm

PR #322, opened 2023-05-10
[DEVHAS-359] use patch instead of update for status update

yangcao77:
### What does this PR do?:

 use patch instead of update for status update

https://issues.redhat.com/browse/DEVHAS-359

### PR acceptance criteria:

- [ ] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact


controllers/component_controller_conditions.go
```diff
@@ -50,7 +59,7 @@ func (r *ComponentReconciler) SetCreateConditionAndUpdateCR(ctx context.Context,
 	}
 	meta.SetStatusCondition(&component.Status.Conditions, condition)
 
-	err := r.Client.Status().Update(ctx, component)
+	err = r.Client.Status().Patch(ctx, component, patch)
```
kim-tsao, line 62: I think we need to pass in the address of the object `component` according to the [api docs](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.14.4/pkg/client/interfaces.go#LL75C7-L75C7).  This is also being done in the ASEB [patches](https://github.com/redhat-appstudio/application-service/blob/main/controllers/applicationsnapshotenvironmentbinding_controller_conditions.go#L61)

yangcao77: the `component` is already passed in as `*appstudiov1alpha1.Component` type

kim-tsao (approved):
/lgtm

controllers/component_controller_conditions.go
```diff
@@ -49,22 +51,21 @@ func (r *ComponentReconciler) SetCreateConditionAndUpdateCR(ctx context.Context,
...
 	err := r.Client.Status().Update(ctx, component)
 	if err != nil {
 		// Retry, and if still fails, then return an error
-		curComponent := &appstudiov1alpha1.Component{}
-		err := r.Client.Get(ctx, req.NamespacedName, curComponent)
+		var currentComponent appstudiov1alpha1.Component
+		err := r.Get(ctx, req.NamespacedName, &currentComponent)
 		if err != nil {
-			return err
+			return nil
```
kim-tsao, line 60: didn't notice this before but did you mean to return nil instead of err?  If so can you update the comment on line 56 which says otherwise?

controllers/component_controller_conditions.go
```diff
@@ -99,18 +100,18 @@ func (r *ComponentReconciler) SetUpdateConditionAndUpdateCR(ctx context.Context,
...
```
kim-tsao, line 106: Same comment as before

controllers/component_controller_conditions.go
```diff
@@ -141,18 +143,21 @@ func (r *ComponentReconciler) SetGitOpsGeneratedConditionAndUpdateCR(ctx context
...
```
kim-tsao, line 152: same comment as before

controllers/component_controller_conditions.go
yangcao77: updated

controllers/component_controller_conditions.go
yangcao77: updated to return error if fail to get component at the retry

controllers/component_controller_conditions.go
yangcao77: updated to return error if fail to get component at the retry

controllers/application_controller_conditions.go
```diff
@@ -24,57 +24,75 @@ import (
...
 )
 
 func (r *ApplicationReconciler) SetCreateConditionAndUpdateCR(ctx context.Context, req ctrl.Request, application *appstudiov1alpha1.Application, createError error) {
 	log := ctrl.LoggerFrom(ctx)
 
+	condition := metav1.Condition{}
+	var currentApplication appstudiov1alpha1.Application
+	err := r.Get(ctx, req.NamespacedName, &currentApplication)
+	if err != nil {
+		return
```
kim-tsao, line 39: should we log an error here too?

controllers/application_controller_conditions.go
```diff
@@ -24,57 +24,75 @@ import (
...
 }
 
 func (r *ApplicationReconciler) SetUpdateConditionAndUpdateCR(ctx context.Context, req ctrl.Request, application *appstudiov1alpha1.Application, updateError error) {
 	log := ctrl.LoggerFrom(ctx)
-
+	condition := metav1.Condition{}
+	var currentApplication appstudiov1alpha1.Application
+	err := r.Get(ctx, req.NamespacedName, &currentApplication)
+	if err != nil {
+		return
```
kim-tsao, line 73: should we log an error here too?

kim-tsao (approved):

controllers/application_controller_conditions.go
yangcao77: updated

controllers/application_controller_conditions.go
yangcao77: updated

PR #328, opened 2023-05-15
[DEVHAS-361] add timeout to cdq reconcile and add url validation

yangcao77:
### What does this PR do?:

add 5 mins timeout to cdq reconcile and add url validation. the validation won't check for status code as private repo may also get a non-200 code. it will be handled by the rest of cdq code later. 

https://issues.redhat.com/browse/DEVHAS-361

### PR acceptance criteria:

- [x] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -28,6 +28,7 @@ import (
 	"reflect"
 	"regexp"
 	"strings"
+	"time"
 
 	appstudiov1alpha1 "github.com/redhat-appstudio/application-api/api/v1alpha1"
 	gitopsgenv1alpha1 "github.com/redhat-developer/gitops-generator/api/v1alpha1"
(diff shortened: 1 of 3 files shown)


pkg/util/util.go
```diff
@@ -164,6 +165,31 @@ func CurlEndpoint(endpoint string) ([]byte, error) {
...
+	u, err := url.Parse(endpoint)
+	if err != nil {
+		return fmt.Errorf("failed to parse the url: %v, err: %v", endpoint, err)
+	}
+
+	if len(u.Host) == 0 || len(u.Scheme) == 0 {
+		return fmt.Errorf("url %v is invalid", endpoint)
+	}
+
+	client := &http.Client{Timeout: 3 * time.Second}
```
kim-tsao, line 181: Will this fail for all private repos since we're not passing in the bearer token?

yangcao77: this validation does not check for status code return. which means, as long as the URL is valid, it will pass. for example, for a private repo `https://github.com/yangcao77/multi-components-private`, it won't return any error for `http.get`, it's reachable, just the status code will be `404` without any token

yangcao77: added private repo url to the unit test to avoid any confusion.

kim-tsao: Thanks for clarifying

kim-tsao (approved):
/lgtm

PR #331, opened 2023-05-16
Add Pact tests and job running them.

Katka92:
### What does this PR do?:
Add pact tests, run the verification against contracts that are already existing in the Pact broker.
Make this as a separate job and make this job as optional, so for now it will not block any PR merge.

https://issues.redhat.com/browse/HAC-2988

To run the pact tests, execute `make pact`.


johnmcollier:
Hi @Katka92

I was looking over this PR and also https://github.com/redhat-appstudio/application-service/pull/336, and I think this PR belongs in [redhat-appstudio/application-api](https://github.com/redhat-appstudio/application-api), rather than HAS? As I’m not seeing anything HAS-specific in the tests?

controllers/application_pact_test_state_handlers.go
```diff
@@ -0,0 +1,91 @@
+//
```
johnmcollier, line 1: Can you add the following files to the ignore list in `codecov.yml`:
- `application_pact_test_state_handlers.go`
- `application_pact_test_utils.go`
- `start_test_env.go`

I think codecov only excludes test files ending in _test by default, so I think the above three test files will also need to be ignored.

controllers/start_test_env.go
```diff
@@ -0,0 +1,138 @@
...
+)
+
+var (
+	k8sClient client.Client // You'll be using this client in your tests.
+	testEnv   *envtest.Environment
+	ctx       context.Context
+	cancel    context.CancelFunc
+)
+
+func setupTestEnv() {
```
johnmcollier, line 52: What's this file needed for compared to just keeping things in `controllers/suite_test.go`?

controllers/application_pact_test_state_handlers.go
Katka92: Thank you, I've excluded those.

johnmcollier (approved):
Looks good 👍

johnmcollier (approved):

PR #334, opened 2023-05-18
[DEVHAS-292] Use tini to manage zombie processes

kim-tsao:
### What does this PR do?:

Install the tini package and update our entrypoint to run under tini 
Reference: https://github.com/krallin/tini#more

https://issues.redhat.com/browse/RHTAPBUGS-292

### PR acceptance criteria:

- [ ] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact


elsony (approved):
The change looks ok and straight forward. The main thing to check on this PR is to fully test the change out after switching to tini to make sure there is no side effect on the change.

maysunfaisal (approved):

Dockerfile
```diff
@@ -33,6 +33,16 @@ COPY --from=builder /workspace/manager .
 COPY appdata.gitconfig /.gitconfig
 RUN chgrp -R 0 /.gitconfig && chmod -R g=u /.gitconfig
 
+# build tini and install to manage zombie proceses
+RUN git clone --branch v0.19.0 https://github.com/krallin/tini /tini
```
johnmcollier, line 37: Perhaps this should be moved to its own step? Otherwise we'll have to rebuild tini each time we modify HAS.

johnmcollier (commented):
Just one suggestion

Dockerfile
kim-tsao: yes, we will need to avoid rebuilding

maysunfaisal (approved):

PR #346, opened 2023-06-15
[RHTAPBUGS-418]multiple components with context path

stuartwdouglas:
Test for [RHTAPBUGS-418](https://issues.redhat.com//browse/RHTAPBUGS-418)

### What does this PR do?:
Test for an issue I saw in the UI where importing a project with multiple components and a context path set results in the context path going missing.

I could not get this test to run locally as all the detection tests are timing out for me.

https://issues.redhat.com/browse/RHTAPBUGS-418

### DIFF
--- a/pkg/devfile/detect.go
+++ b/pkg/devfile/detect.go
@@ -47,7 +47,6 @@ type AlizerClient struct {
 // Map 3 returns a context to the Dockerfile uri or a matched DockerfileURL from the devfile registry if no Dockerfile is present in the context
 // Map 4 returns a context to the list of ports that were detected by alizer in the source code, at that given context
 func search(log logr.Logger, a Alizer, localpath string, devfileRegistryURL string, source appstudiov1alpha1.GitSource) (map[string][]byte, map[string]string, map[string]string, map[string][]int, error) {
-
 	devfileMapFromRepo := make(map[string][]byte)
 	devfilesURLMapFromRepo := make(map[string]string)
 	dockerfileContextMapFromRepo := make(map[string]string)
@@ -63,7 +62,8 @@ func search(log logr.Logger, a Alizer, localpath string, devfileRegistryURL stri
 			isDevfilePresent := false
 			isDockerfilePresent := false
 			curPath := path.Join(localpath, f.Name())
-			context := f.Name()
+			dirName := f.Name()
+			context := path.Join(source.Context, f.Name())
 			files, err := ioutil.ReadDir(curPath)
 			if err != nil {
 				return nil, nil, nil, nil, err
@@ -74,7 +74,7 @@ func search(log logr.Logger, a Alizer, localpath string, devfileRegistryURL stri
 					/* #nosec G304 -- false positive, filename is not based on user input*/
 					devfilePath := path.Join(curPath, f.Name())
 					// Set the proper devfile URL for the detected devfile
-					updatedLink, err := UpdateGitLink(source.URL, source.Revision, path.Join(source.Context, path.Join(context, f.Name())))
+					updatedLink, err := UpdateGitLink(source.URL, source.Revision, path.Join(source.Context, path.Join(dirName, f.Name())))
 					if err != nil {
 						return nil, nil, nil, nil, err
 					}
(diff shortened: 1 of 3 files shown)


pkg/devfile/devfile.go
```diff
@@ -851,7 +851,28 @@ func DownloadDevfileAndDockerfile(url string) ([]byte, string, []byte, string) {
 // Map 3 returns a context to the Dockerfile uri or a matched DockerfileURL from the devfile registry if no Dockerfile/Containerfile is present in the context
 // Map 4 returns a context to the list of ports that were detected by alizer in the source code, at that given context
 func ScanRepo(log logr.Logger, a Alizer, localpath string, devfileRegistryURL string, source appstudiov1alpha1.GitSource) (map[string][]byte, map[string]string, map[string]string, map[string][]int, error) {
```
stuartwdouglas, line 853: I am not sure if the original method could still be used anywhere else, so I added a new one instead

stuartwdouglas: I have updated the PR

stuartwdouglas:
Does this need any more action on my part?

controllers/componentdetectionquery_controller.go
```diff
@@ -284,7 +284,7 @@ func (r *ComponentDetectionQueryReconciler) Reconcile(ctx context.Context, req c
 			if isMultiComponent {
 				log.Info(fmt.Sprintf("Since this is a multi-component, attempt will be made to read only level 1 dir for devfiles... %v", req.NamespacedName))
 
-				devfilesMap, devfilesURLMap, dockerfileContextMap, componentPortsMap, err = devfile.ScanRepo(log, r.AlizerClient, componentPath, r.DevfileRegistryURL, source)
+				devfilesMap, devfilesURLMap, dockerfileContextMap, componentPortsMap, err = devfile.ScanRepo(log, r.AlizerClient, source.Context, componentPath, r.DevfileRegistryURL, source)
```
yangcao77, line 287: the source is already been passed in as a arg. no need an extra arg for `source.Context`

stuartwdouglas: oops, fixed

pkg/devfile/devfile.go
```diff
@@ -851,7 +853,15 @@ func DownloadDevfileAndDockerfile(url string) ([]byte, string, []byte, string) {
...
-	return search(log, a, localpath, devfileRegistryURL, source)
+	originalContext := source.Context
```
yangcao77, line 856: can this block be moved to `search`?  
just trying to avoid the API arg change.

yangcao77 (changes requested):

pkg/devfile/devfile.go
yangcao77: actually, this code block can be removed. the `source.context` can be get in `search`

pkg/devfile/detect.go
```diff
@@ -63,7 +63,8 @@ func search(log logr.Logger, a Alizer, localpath string, devfileRegistryURL stri
 			isDevfilePresent := false
 			isDockerfilePresent := false
 			curPath := path.Join(localpath, f.Name())
-			context := f.Name()
+			dirName := f.Name()
+			context := originalContext + f.Name()
```
yangcao77, line 67: context := path.Join(source.Context, f.Name())

pkg/devfile/devfile.go
yangcao77: see https://github.com/redhat-appstudio/application-service/pull/346/files#r1255764499

stuartwdouglas:
Fixed

pkg/devfile/detect.go
```diff
@@ -47,7 +47,14 @@ type AlizerClient struct {
...
 func search(log logr.Logger, a Alizer, localpath string, devfileRegistryURL string, source appstudiov1alpha1.GitSource) (map[string][]byte, map[string]string, map[string]string, map[string][]int, error) {
-
+	originalContext := source.Context
+	if originalContext != "" {
+		if originalContext == "./" {
+			originalContext = ""
+		} else if !strings.HasSuffix(originalContext, "/") {
+			originalContext += "/"
+		}
+	}
```
yangcao77, line 57: do not need this code change.  the context will never be empty. as it's already set in https://github.com/redhat-appstudio/application-service/blob/eeef42302b0bfcac2877973878966f0e880b7fe7/controllers/componentdetectionquery_controller.go#L147-L149

and with the `path.join`, you do not need to append `/` at the end

yangcao77 (changes requested):

yangcao77 (approved):

PR #363, opened 2023-07-12
[DEVHAS-380] Have the Application controller add/remove components from its model

johnmcollier:
### What does this PR do?:
This PR updates the logic that handles the Application's devfile model by moving the code in the Component controller that adds/removes components from it, and moves it to the Application controller. 

With these changes:
- Application reconcile is triggered by a Component creation or deletion
- During each reconcile of an Application, the Application controller retrieves the list of Components that set the Application as its owner, and updates its devfile model accordingly

https://issues.redhat.com/browse/DEVHAS-380

### PR acceptance criteria:

- [ ] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact


controllers/application_controller.go
```diff
@@ -196,11 +209,24 @@ func (r *ApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 			return ctrl.Result{}, err
 		}
 
+		updateRequired := false
+		// nil out the attributes and projects for the application devfile
+		// The Attributes contain any image components for the application
+		// And the projects contains any git components for the application
+		devWorkspacesSpec := devfileData.GetDevfileWorkspaceSpec().DeepCopy()
+		devWorkspacesSpec.Attributes = nil
+		devWorkspacesSpec.Projects = nil
```
maysunfaisal, line 218: wait why do we do this again

controllers/mapper.go
```diff
@@ -63,3 +63,23 @@ func MapToBindingByBoundObjectName(cl client.Client, objectType, label string) f
...
+			return []reconcile.Request{
+				{
+					NamespacedName: types.NamespacedName{
+						Namespace: component.Namespace,
+						Name:      component.Spec.Application,
+					},
+				},
+			}
+		}
+		// the obj was not a namespace or it did not have the required label.
```
maysunfaisal, line 82:
```suggestion
		// the obj was not in the namespace or it did not have the required Application.
```
should be?

controllers/application_controller.go
johnmcollier: To make comparisons of the added git projects (under projects in the devfile spec) and container image projects (under attributes) easier in the devfile app model. Basically we create a copy of the devworkspace spec, nil out the attributes and projects, add the components to it, and compare with the existing devworkspace spec.

maysunfaisal (approved):

johnmcollier:
Need to wait for pact tests to be updated before merging

/hold

yangcao77 (approved):

PR #364, opened 2023-07-14
[DEVHAS-421] make git clone and cdq analysis a separate go module

yangcao77:
### What does this PR do?:

move git clone and cdq analysis out of cdq controller to a separate go module

https://issues.redhat.com/browse/DEVHAS-421

### PR acceptance criteria:

- [x] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact


cdq-analysis/main.go
```diff
@@ -0,0 +1,79 @@
...
+			fmt.Printf("Error creating clientset with config %v: %v", config, err)
+		}
+	}
+	k8sInfoClient := pkg.K8sInfoClient{
+		Ctx:          ctx,
+		Clientset:    clientset,
+		Log:          log,
+		CreateK8sJob: createK8sJob,
+	}
+	pkg.CloneAndAnalyze(k8sInfoClient, gitToken, namespace, name, contextPath, devfilePath, dockerfilePath, URL, Revision, DevfileRegistryURL, isDevfilePresent, isDockerfilePresent)
```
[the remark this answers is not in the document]
yangcao77: the main.go is trigger by docker image, and the result will be send by the k8s job. no need return anything from here.

cdq-analysis/pkg/util.go
```diff
@@ -0,0 +1,218 @@
...
+	"github.com/devfile/registry-support/index/generator/schema"
+	registryLibrary "github.com/devfile/registry-support/registry-library/library"
+	"github.com/redhat-developer/alizer/go/pkg/apis/model"
+)
+
+// CloneRepo clones the repoURL to clonePath
+func CloneRepo(clonePath, repoURL string, revision string, token string) error {
+	exist, err := IsExist(clonePath)
+	if !exist || err != nil {
+		os.MkdirAll(clonePath, 0750)
```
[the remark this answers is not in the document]
yangcao77: it was there from previous code.

cdq-analysis/main.go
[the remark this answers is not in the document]
maysunfaisal: will this module reside in this repository?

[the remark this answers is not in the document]
yangcao77: yes, this module will be in this repo.

cdq-analysis/main.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+func main() {
+	// remove the prefix and suffix quotes
+	for i := 1; i <= 10; i++ {
+		if strings.HasPrefix(os.Args[i], "\"") && strings.HasSuffix(os.Args[i], "\"") && len(os.Args[i]) > 1 {
+			os.Args[i] = os.Args[i][1 : len(os.Args[i])-1]
+		}
+	}
+	gitToken := os.Args[1]
```
johnmcollier, line 40: Do we need validation of the command-line args?

cdq-analysis/main.go
```diff
@@ -0,0 +1,79 @@
...
+	createK8sJob, _ := strconv.ParseBool(os.Args[12])
+
+	opts := zap.Options{
+		TimeEncoder: zapcore.ISO8601TimeEncoder,
+	}
+	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
+	log := ctrl.Log.WithName("cdq-analysis").WithName("CloneAndAnalyze")
+	var ctx context.Context
+	var clientset *kubernetes.Clientset
+	if createK8sJob {
```
johnmcollier, line 60: Should this flag instead be outside of the module and be part of HAS? Because HAS will be what's launching the Kubernetes Job containing this go module (and main.go)

FWIW, the way I split between using the go module vs. the job for gitops generation was like this: https://github.com/redhat-appstudio/application-service/blob/d24a233c36d89456366207da5a2438498954cb3e/controllers/component_controller.go#L532-L544. Basically if LOCAL_GITOPS was set, or an annotation on the CR was set, then we would invoke the Go module directly, otherwise we would launch the Kubernetes job containing the module

johnmcollier: Sorry, what I mean is, is that the logic that splits between whether or not to launch a job or to launch a module, should reside in HAS, rather than in the Go module.

cdq-analysis/main.go
yangcao77: the main.go is actually not being used by HAS controller yet. similarly, will be part of the docker image issue. may I add validation as part of that issue?

cdq-analysis/main.go
yangcao77: yes. I understands that, but the go module will need that env as well. since it need to determine if a configmap need to be created when returning the result.  
in this PR, we are using go function to return the result. 
the job not yet created. and will have that ENV been set in HAS when a job is ready to be consumed.

cdq-analysis/main.go
johnmcollier: Sure

johnmcollier (approved):
Looks fine from my end

PR #370, opened 2023-07-26
[DEVHAS-422]add docker image build to ci

yangcao77:
### What does this PR do?:

Switch to use command flags with entrypoint.sh to execute main.go. added flags validation

Build the docker image in PR build. 

added tekton pipeline to push image upon cdq-analysis path change.  
to build and monitor the change for a particular context, got reference from [qulity-dashboard](https://github.com/redhat-appstudio/quality-dashboard/blob/main/.tekton/pull-request-backend.yaml)

https://issues.redhat.com/browse/DEVHAS-422

### PR acceptance criteria:

- [ ] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact


.github/workflows/pr.yml
```diff
@@ -121,6 +121,22 @@ jobs:
...
+        with:
+          fetch-depth: 0
+      - name: Set up Go 1.x
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.18
+      - name: Change to the cdq-analysis directory
+        run: cd cdq-analysis/
+      - name: Check if dockerimage build is working
+        run: docker build -f ./Dockerfile .
```
johnmcollier, line 139: We should use the same docker buildx build that the workflow on pushes to main uses, so that the builds on PR checks and post-merge are consistent

cdq-analysis/Dockerfile
```diff
@@ -6,5 +6,9 @@ COPY  . .
 
 RUN go mod download
 
-RUN go build -o ./main
+RUN CGO_ENABLED=0 GOOS=linux go build -a -o cdq-analysis main.go
```
johnmcollier, line 9: It's probably better to use a separate builder stage and then use a smaller base image (like ubi minimal) for the runtime stage E.g.:

https://github.com/redhat-appstudio/application-service/blob/main/Dockerfile#L32

yangcao77: updated

.github/workflows/pr.yml
yangcao77: but the post-merge action is to push the image to dockerhub, and we do not want to push images to dockerhun for PR builds, right?

johnmcollier: Hmm, didn't noticed the build _and_ push in that step. I think copying over the docker buildx setup step should be sufficient then, and then just keeping the `docker build -f ./Dockerfile`

cdq-analysis/Dockerfile
```diff
@@ -1,10 +1,43 @@
-FROM registry.access.redhat.com/ubi9/go-toolset:latest
+# FROM registry.access.redhat.com/ubi9/go-toolset:latest
+
+# WORKDIR /app
+
+# COPY  . .
+
+# RUN go mod download
+
+# RUN CGO_ENABLED=0 GOOS=linux go build -a -o cdq-analysis main.go
```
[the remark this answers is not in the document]
yangcao77: updated

cdq-analysis/main.go
```diff
@@ -31,24 +31,30 @@ import (
...
+	flag.StringVar(&Revision, "revision", "", "The revision of the git repo to run cdq analysis against with")
+	flag.StringVar(&DevfileRegistryURL, "devfileRegistryURL", pkg.DevfileRegistryEndpoint, "The devfile registry URL")
+	flag.StringVar(&namespace, "namespace", "", "The namespace from which to fetch resources")
+	flag.BoolVar(&isDevfilePresent, "isDevfilePresent", false, "If the devfile present in the root of the repository")
+	flag.BoolVar(&isDockerfilePresent, "isDockerfilePresent", false, "If the dockerfile present in the root of the repository")
+	flag.BoolVar(&createK8sJob, "createK8sJob", false, "If a kubernetes job need to be created to send back the result")
+	flag.Parse()
+
+	if err := validateVariables(name, URL, namespace, Revision); err != nil {
+		log.Fatal(err)
```
maysunfaisal, line 56: do we want to return here?

johnmcollier (approved):
Looks fine from my end

PR #379, opened 2023-08-14
Provide a Service Port Name

maysunfaisal:
### What does this PR do?:

Provides a Service Port name if there is more than one Service Port. The API says Name is mandatory is there is more than 1.

https://issues.redhat.com/browse/RHTAPBUGS-433

### PR acceptance criteria:

- [x] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact

  

Create a Component, SEB, Snapshot, Environment. Ensure the Gitops Repo service.yaml has two service ports and both of them have names. Apply it in your OpenShift project and access the route.

Example of the Gitops Repo for providing Service Port Name https://github.com/maysunfaisal/application-sample-lK5Sb-borrow-place/blob/main/components/component-sample/base/service.yaml#L13-L20

### DIFF
--- a/pkg/devfile/devfile.go
+++ b/pkg/devfile/devfile.go
@@ -341,20 +341,32 @@ func GetResourceFromDevfile(log logr.Logger, devfileData data.DevfileData, deplo
 
 					if currentPort > 0 {
 						servicePort := corev1.ServicePort{
+							Name:       strconv.Itoa(currentPort),
 							Port:       int32(currentPort),
 							TargetPort: intstr.FromInt(currentPort),
 						}
 
 						isPresent := false
+						portNameMap := make(map[string]bool)
 						for _, port := range resources.Services[0].Spec.Ports {
+							portNameMap[port.Name] = true
 							if port.Port == servicePort.Port {
 								isPresent = true
-								break
 							}
 						}
 
 						if !isPresent {
+							if portNameMap[servicePort.Name] {
+								servicePort.Name = fmt.Sprintf("%s-%s", servicePort.Name, util.GetRandomString(4, true))
+							}
 							resources.Services[0].Spec.Ports = append(resources.Services[0].Spec.Ports, servicePort)
+							for i, port := range resources.Services[0].Spec.Ports {
+								if port.Name == "" {
+									// if port name is empty for other service ports, assign a name
+									// because name is required if there is more than one port
+									resources.Services[0].Spec.Ports[i].Name = strconv.Itoa(int(port.Port))
+								}
+							}
 						}
 					}
 				}
--- a/pkg/devfile/devfile.go
+++ b/pkg/devfile/devfile.go
@@ -357,14 +357,22 @@ func GetResourceFromDevfile(log logr.Logger, devfileData data.DevfileData, deplo
 
 						if !isPresent {
 							if portNameMap[servicePort.Name] {
-								servicePort.Name = fmt.Sprintf("%s-%s", servicePort.Name, util.GetRandomString(4, true))
+								generatedName := fmt.Sprintf("%s-%s", servicePort.Name, util.GetRandomString(4, true))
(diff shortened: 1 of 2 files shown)


pkg/devfile/devfile.go
```diff
@@ -341,20 +341,40 @@ func GetResourceFromDevfile(log logr.Logger, devfileData data.DevfileData, deplo
 
 					if currentPort > 0 {
 						servicePort := corev1.ServicePort{
+							Name:       strconv.Itoa(currentPort),
```
yangcao77, line 344: I kinda remember the service port name cannot start with a number. so perhaps use `fmt.Sprintf(port-%v),currentPort)` instead?

pkg/devfile/devfile.go
```diff
@@ -341,20 +341,40 @@ func GetResourceFromDevfile(log logr.Logger, devfileData data.DevfileData, deplo
...
+							}
 							resources.Services[0].Spec.Ports = append(resources.Services[0].Spec.Ports, servicePort)
+
+							for i, port := range resources.Services[0].Spec.Ports {
+								if port.Name == "" {
+									// if port name is empty for other service ports, assign a name
+									// because name is required if there is more than one port
+									portName := strconv.Itoa(int(port.Port))
+									if portNameMap[portName] {
+										portName = fmt.Sprintf("%s-%s", portName, util.GetRandomString(4, true))
```
yangcao77, line 372: I'm wondering what is the purpose of having multiple service ports with different names but same port number? I know it's not blocked by kube api, but it's just exposing the same port to be accessible, isn't it?

pkg/devfile/devfile.go
maysunfaisal: I found it https://kubernetes.io/docs/concepts/services-networking/service/

Screenshot 2023-08-15 at 3 25 10 PM

maysunfaisal: they can start and end with alphanumeric lowercase

pkg/devfile/devfile.go
maysunfaisal: Good question.. I was having the exact same questions when i was writing this 😆 
but I was thinking, we cant control what ppl name their ports, if someone is bring in a service with port 8080 and naming it "1234", we cant do anything about it.

However, if they enter 1234 as port on RHTAP HAC, then we cant use "1234" as name anymore. This is more for avoiding name clashes. Any other better suggestions?

yangcao77: makes sense.

pkg/devfile/devfile.go
```diff
@@ -341,20 +341,40 @@ func GetResourceFromDevfile(log logr.Logger, devfileData data.DevfileData, deplo
...
 
 						if !isPresent {
+							if portNameMap[servicePort.Name] {
+								generatedName := fmt.Sprintf("%s-%s", servicePort.Name, util.GetRandomString(4, true))
+								portNameMap[generatedName] = true
+								servicePort.Name = generatedName
+							}
 							resources.Services[0].Spec.Ports = append(resources.Services[0].Spec.Ports, servicePort)
+
+							for i, port := range resources.Services[0].Spec.Ports {
```
yangcao77, line 366: any reason we are iterating through the ports twice?

yangcao77: sounds reasonable to me, 
just one additional thing,  since the first loop is just to check the existence. if the port ever present, we should break out of the loop.

yangcao77 (approved):
generally looks good to me, just a small question as commented.

pkg/devfile/devfile.go
maysunfaisal: We dont break because I need to populate `portNameMap[port.Name] = true` on first loop

Then use the map to see if our Component spec port name needs to be updated, when appending

PR #382, opened 2023-08-21
[DEVHAS-418]add dockerfile as an alternate name for Dockerfile

yangcao77:
### What does this PR do?:

This PR adds `dockerfile` as an alternate name for `Dockerfile`

also added unit test for it, and update existing test repo for the controller tests. 

https://issues.redhat.com/browse/DEVHAS-418

### PR acceptance criteria:

- [x] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact

### DIFF
--- a/cdq-analysis/pkg/detect.go
+++ b/cdq-analysis/pkg/detect.go
@@ -68,7 +68,8 @@ func search(log logr.Logger, a Alizer, localpath string, devfileRegistryURL stri
 				return nil, nil, nil, nil, err
 			}
 			for _, f := range files {
-				if f.Name() == DevfileName || f.Name() == HiddenDevfileName {
+				lowerCaseFileName := strings.ToLower(f.Name())
+				if lowerCaseFileName == DevfileName || lowerCaseFileName == HiddenDevfileName {
 					// Check for devfile.yaml or .devfile.yaml
 					/* #nosec G304 -- false positive, filename is not based on user input*/
 					devfilePath := path.Join(curPath, f.Name())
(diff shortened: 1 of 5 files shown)


cdq-analysis/pkg/detect.go
```diff
@@ -122,21 +122,22 @@ func search(log logr.Logger, a Alizer, localpath string, devfileRegistryURL stri
 							}
 						}
 					}
-				} else if f.Name() == DockerfileName {
-					// Check for Dockerfile
+				} else if f.Name() == DockerfileName || f.Name() == AlternateDockerfileName {
```
johnmcollier, line 125: Small suggestion: Rather than checking for alternate types of Dockerfile names, what if we convert the name to lower case and check against "dockerfile"?

yangcao77: then any other mix-case names are also allowed? i.e. `DoERerFiLE`?

yangcao77: I think we still need to check the exact allowed name, due to https://github.com/redhat-appstudio/application-service/blob/902d8977998183611795a8d5eae5ead8aee04e95/pkg/devfile/devfile.go#L750-L756

When looking for the files using endpoint, we still need to check for supported names. 

But in the cdq analysis here, when we have the repo downloaded on disk, I can convert to lower case and check against that.

yangcao77: updated

maysunfaisal (approved):

PR #383, opened 2023-08-23
Add metric for Application deletion

johnmcollier:
### What does this PR do?:
Adds a metric for tracking successful and failed application deletions in HAS.

https://issues.redhat.com/browse/DEVHAS-452

### PR acceptance criteria:

- [ ] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact

### DIFF
--- a/controllers/application_controller.go
+++ b/controllers/application_controller.go
@@ -111,6 +111,7 @@ func (r *ApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		}
 	} else {
 		if containsString(application.GetFinalizers(), appFinalizerName) {
+			metrics.ApplicationDeletionTotalReqs.Inc()
 			// A finalizer is present for the Application CR, so make sure we do the necessary cleanup steps
 			if err := r.Finalize(ctx, &application, ghClient); err != nil {
 				finalizeCounter, err := getCounterAnnotation(finalizeCount, &application)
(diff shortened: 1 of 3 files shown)


controllers/application_controller.go
```diff
@@ -126,6 +127,9 @@ func (r *ApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 					// if fail to delete the external dependency here, log the error, but don't return error
 					// Don't want to get stuck in a cycle of repeatedly trying to delete the repository and failing
 					log.Error(err, "Unable to delete GitOps repository for application %v in namespace %v", application.GetName(), application.GetNamespace())
+
+					// Increment the Application deletion failed metric as the application delete did not fully succeed
+					metrics.ApplicationDeletionFailed.Inc()
```
kim-tsao, line 132: Are we going to count the [parse error](https://github.com/redhat-appstudio/application-service/blob/main/controllers/application_controller_finalizer.go#L56) as part of the SLO?  Will the finalizer get called if the Application CR was not created properly due to bad input?

johnmcollier: I’ll update to count the parse error.

Bad input won’t prevent the finalizer from executing

kim-tsao: right, which means we shouldn't be counting the parse error or the error if the gitops repo URL attribute is not found?

kim-tsao (approved):

PR #388, opened 2023-09-05
[DEVHAS183] create kube jobs for CDQ analysis

yangcao77:
### What does this PR do?:

This PR spawns Kube jobs for CDQ analysis using the cdq-analysis image. 
sets a environment variable and also checks for annotation to decide if the cdq analysis should be run using go module or Kube job. The default is still running the go module. 
The results and errors are returned by configmap.

also creates functional tests for the kube job

https://issues.redhat.com/browse/DEVHAS-183

### PR acceptance criteria:

- [x] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact


controllers/componentdetectionquery_controller.go
```diff
@@ -157,32 +167,144 @@ func (r *ComponentDetectionQueryReconciler) Reconcile(ctx context.Context, req c
 
 		if source.DevfileURL == "" {
 			log.Info(fmt.Sprintf("Attempting to read a devfile from the URL %s... %v", source.URL, req.NamespacedName))
+			var devfilesMapReturned map[string][]byte
+			var devfilesURLMapReturned, dockerfileContextMapReturned map[string]string
+			var componentPortsMapReturned map[string][]int
+			revision := source.Revision
+
+			if r.RunKubernetesJob && !(componentDetectionQuery.Annotations["runCDQAnalysisLocal"] == "true") {
```
kim-tsao, line 175: I think we should explain the use of the `runCDQAnalysisLocal` annotation somewhere.  Maybe add a comment here and in the CONTRIBUTING.MD to explain it's used as a flag to run tests in a go module?

controllers/start_test_env.go
```diff
@@ -122,6 +122,9 @@ func setupTestEnv() {
 		GitHubTokenClient:  mockGhTokenClient,
 		DevfileRegistryURL: cdqanalysis.DevfileStageRegistryEndpoint, // Use the staging devfile registry for tests
 		AppFS:              ioutils.NewMemoryFilesystem(),
+		Config:             cfg,
+		RunKubernetesJob:   true,
+		CdqAnalysisImage:   "quay.io/redhat-appstudio/cdq-analysis:next",
```
kim-tsao, line 127: Should we update the readme test instructions?  If someone makes changes in the cdq-analysis code, they will have to rebuild their own image and add it here for testing

kim-tsao: Or maybe have users set an environment variable that defaults to `quay.io/redhat-appstudio/cdq-analysis:next` if unset

controllers/componentdetectionquery_controller.go
```diff
@@ -157,32 +167,144 @@ func (r *ComponentDetectionQueryReconciler) Reconcile(ctx context.Context, req c
...
+				} else {
+					//print job details
+					log.Info(fmt.Sprintf("Successfully created cdq analysis job %v, waiting for config map to be created... %v", jobName, req.NamespacedName))
+				}
 
-			k8sInfoClient := cdqanalysis.K8sInfoClient{
-				Log:          log,
-				CreateK8sJob: false,
-			}
+				cm, err := waitForConfigMap(clientset, ctx, req.Name, req.Namespace)
```
kim-tsao, line 251: The timeout to wait for the configMap is 5mins but is it generally a lot quicker than that? I see that it's constructed and created in the CDQ module so most of the processing is done - any delays would be attributed to the reading of the configMap from the cluster.

controllers/componentdetectionquery_controller.go
```diff
@@ -184,18 +195,149 @@
...
-				return ctrl.Result{}, nil
+				cm, err := waitForConfigMap(clientset, ctx, req.Name, req.Namespace)
+				if err != nil {
+					log.Error(err, fmt.Sprintf("Error waiting for configmap creation ...%v", req.NamespacedName))
+					r.SetCompleteConditionAndUpdateCR(ctx, req, &componentDetectionQuery, copiedCDQ, err)
+					cleanupK8sResources(log, clientset, ctx, fmt.Sprintf("%s-job", req.Name), req.Name, req.Namespace)
+					return ctrl.Result{}, nil
+				}
+				var errMapReturned map[string]string
+				json.Unmarshal(cm.BinaryData["devfilesMap"], &devfilesMapReturned)
```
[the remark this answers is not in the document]
kim-tsao: Can we handle all json unmarshalling errors?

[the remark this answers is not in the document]
yangcao77: we are not expecting to receive all those data. if any map is empty or the returnErr map is nil, it will result in unmarshal error as well. So I can only log those error instead of return Reconcile immediately
```
"5 errors occurred:\n\t* unmarshal devfilesMap: unexpected end of JSON input\n\t* unmarshal dockerfileContextMap: unexpected end of JSON input\n\t* unmarshal devfilesURLMap: unexpected end of JSON input\n\t* unmarshal componentPortsMap: unexpected end of JSON input\n\t* unmarshal revision: unexpected end of JSON input\n\n"
```

controllers/componentdetectionquery_controller.go
yangcao77: updated in the description.  thinking if the info is too detailed to be in CONTRIBUTING.MD.

controllers/start_test_env.go
yangcao77: updated

controllers/componentdetectionquery_controller.go
yangcao77: yes, generally the CDQ is quicker, with the cdq-image cached on cluster, the entire CDQ time only takes a few seconds. and yes, at this point,most of the processing is done, so we are expecting to use the waiting time as the CDQ timeout time. The reconcile context timeout does not work as expected. it will wait for the sub-process to be done before it checks the timeout. so by switching to use the config map waiting time, we can easily cut the CDQ progress upon timeout and make it under control.

yangcao77: and the design for UI side CDQ timeout will be 1min, after 1min, CDQ will still run in backend. we will terminate CDQ if it stuck at waiting for CDQ job to be returned for over 5mins

kim-tsao (approved):
/lgtm

PR #395, opened 2023-10-02
[DEVHAS-482] add metrics git import

yangcao77:
### What does this PR do?:

This PR adds metrics to system errors for importing from git repositories. Added in CDQ and component controller import case

Also adds tests for success and failure case. 

https://issues.redhat.com/browse/DEVHAS-482

https://issues.redhat.com/browse/DEVHAS-524

### PR acceptance criteria:

- [ ] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact


controllers/component_controller.go
```diff
@@ -312,10 +314,15 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
 					if err != nil {
+						// Increment the import git repo failed metric on non-user errors
+						if _, ok := err.(*devfile.NoFileFound); !ok {
+							metrics.ImportGitRepoFailed.Inc()
+						}
 						log.Error(err, fmt.Sprintf("Unable to read the devfile from dir %s %v", gitURL, req.NamespacedName))
 						_ = r.SetCreateConditionAndUpdateCR(ctx, req, &component, err)
 						return ctrl.Result{}, err
 					}
+					metrics.ImportGitRepoSucceeded.Inc()
```
johnmcollier, line 325: Since a "success" scenario here also includes "user errors" (basically anything other than a system error, we also want to increment above when err != nil and the err is `devfile.NoFileFound`.

controllers/component_controller.go
```diff
@@ -324,10 +331,15 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
 					if err != nil {
+						// Increment the import git repo failed metric on non-user errors
+						if _, ok := err.(*devfile.NoFileFound); !ok {
+							metrics.ImportGitRepoFailed.Inc()
+						}
 						log.Error(err, fmt.Sprintf("Unable to download from any known devfile locations from %s %v", gitURL, req.NamespacedName))
 						_ = r.SetCreateConditionAndUpdateCR(ctx, req, &component, err)
 						return ctrl.Result{}, err
 					}
+					metrics.ImportGitRepoSucceeded.Inc()
```
johnmcollier, line 342: Same as above, we also want to increment the success metric in user err scenarios

controllers/component_controller_test.go
```diff
@@ -1028,6 +1040,70 @@ var _ = Describe("Component controller", func() {
...
+
+			// Make sure the devfile model was properly set in Component
+			errCondition := createdHasComp.Status.Conditions[len(createdHasComp.Status.Conditions)-1]
+			Expect(errCondition.Status).Should(Equal(metav1.ConditionFalse))
+
+			hasAppLookupKey := types.NamespacedName{Name: applicationName, Namespace: HASAppNamespace}
+
+			Expect(testutil.ToFloat64(metrics.ImportGitRepoTotalReqs) > beforeImportGitRepoTotalReqs).To(BeTrue())
+			Expect(testutil.ToFloat64(metrics.ImportGitRepoSucceeded) == beforeImportGitRepoSucceeded).To(BeTrue())
+			Expect(testutil.ToFloat64(metrics.ImportGitRepoFailed) > beforeImportGitRepoFailed).To(BeTrue())
```
johnmcollier, line 1097: Why would we want the failed metric to increase here? Since an invalid repo was provided, wouldn't this be considered a user error?

controllers/componentdetectionquery_controller.go
```diff
@@ -320,11 +324,22 @@ func (r *ComponentDetectionQueryReconciler) Reconcile(ctx context.Context, req c
 
 				devfilesMapReturned, devfilesURLMapReturned, dockerfileContextMapReturned, componentPortsMapReturned, revision, err = cdqanalysis.CloneAndAnalyze(k8sInfoClient, req.Namespace, req.Name, context, cdqInfo)
 				if err != nil {
-					log.Error(err, fmt.Sprintf("Error running cdq analysis... %v", req.NamespacedName))
+					switch err.(type) {
+					case *cdqanalysis.NoDevfileFound:
+						log.Error(err, fmt.Sprintf("NoDevfileFound error running cdq analysis... %v", req.NamespacedName))
+					case *cdqanalysis.NoDockerfileFound:
+						log.Error(err, fmt.Sprintf("NoDockerfileFound error running cdq analysis... %v", req.NamespacedName))
+					default:
```
johnmcollier, line 332: Are there any other user errors we need to consider here? What if the repo contains a devfile, but it's invalid or unreadable?

controllers/componentdetectionquery_controller.go
```diff
@@ -320,11 +324,22 @@ func (r *ComponentDetectionQueryReconciler) Reconcile(ctx context.Context, req c
...
+						// Increment the git import failure metric only on non user failure
+						metrics.ImportGitRepoFailed.Inc()
+						log.Error(err, fmt.Sprintf("Internal error running cdq analysis... %v", req.NamespacedName))
+					}
+
 					r.SetCompleteConditionAndUpdateCR(ctx, req, &componentDetectionQuery, copiedCDQ, err)
 					return ctrl.Result{}, nil
 				}
 			}
+			metrics.ImportGitRepoSucceeded.Inc()
```
johnmcollier, line 342: As mentioned above, we should still be incrementing the success counter even in cases of user errors

johnmcollier (changes requested):

controllers/componentdetectionquery_controller.go
yangcao77: add more user cases

controllers/component_controller_test.go
yangcao77: updated

controllers/component_controller.go
yangcao77: updated

controllers/componentdetectionquery_controller.go
yangcao77: updated

controllers/component_controller.go
yangcao77: updated

controllers/componentdetectionquery_controller.go
johnmcollier: Do we need to consider this scenario too? https://github.com/redhat-appstudio/application-service/blob/15f897cd79f7df90e5247c89d7f3438196a3f5c6/cdq-analysis/pkg/componentdetectionquery.go#L197-L202

I tried to walk through Alizer, and it seemed there were a couple cases where it would return errors that would count as user errors (e.g. https://github.com/devfile/alizer/blob/main/pkg/apis/recognizer/component_recognizer.go#L288)

yangcao77: for Alizer, we are calling `DetectComponentsWithSettings` so which calls `getComponentsWithoutConfigFile`, which won't return error if no components being detected
https://github.com/devfile/alizer/blob/6b0b633dc3423413e8677b9ac7de0d778020c9f8/pkg/apis/recognizer/component_recognizer.go#L117C9-L117C9

johnmcollier (approved):

yangcao77:
@johnmcollier  May I get another lgtm label? just pushed the change to fix the test repo in unit test. Verified the tests now passed in https://github.com/redhat-appstudio/application-service/pull/396

johnmcollier (approved):

PR #401, opened 2023-10-23
[DEVHAS-528] Add validating webhook for build-nudges-ref

johnmcollier:
### What does this PR do?:
This PR adds changes to the Component's validating webhook to handle the `build-nudges-ref` field:
- If any nudged components belong to a different application, an error is returned and the create/update operation is blocked
- If any nudged components refer to the nudging component, either directly or indirectly, an error is returned and the operation is blocked

This PR also adds back the stub `Default()` functions for the Application and Component webhooks, as they were accidentally removed when migrating the webhooks from application-api. These functions will be needed to implement mutating webhooks as part of [DEVHAS-529](https://issues.redhat.com/browse/DEVHAS-529)
 
https://issues.redhat.com/browse/DEVHAS-528

### PR acceptance criteria:

- [x] Unit/Functional tests

  

- [x] Documentation 

N/A

- [x] Client Impact

N/A

### DIFF
--- a/controllers/webhooks/component_webhook.go
+++ b/controllers/webhooks/component_webhook.go
@@ -26,7 +26,9 @@ import (
 	appstudiov1alpha1 "github.com/redhat-appstudio/application-api/api/v1alpha1"
 
 	"github.com/go-logr/logr"
+	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/validation"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -44,6 +46,7 @@ func (w *ComponentWebhook) Register(mgr ctrl.Manager, log *logr.Logger) error {
 
 	return ctrl.NewWebhookManagedBy(mgr).
 		For(&appstudiov1alpha1.Component{}).
+		WithDefaulter(w).
 		WithValidator(w).
 		Complete()
 }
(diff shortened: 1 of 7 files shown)


controllers/webhooks/component_webhook.go
```diff
@@ -118,3 +141,33 @@ func (r *ComponentWebhook) ValidateDelete(ctx context.Context, obj runtime.Objec
 	// TODO(user): fill in your validation logic upon object deletion.
 	return nil
 }
+
+// validateBuildNudgesRefTree returns an error if a cycle was found in the 'build-nudges-ref' dependency graph
```
yangcao77, line 145:
```suggestion
// validateBuildNudgesRefGraph returns an error if a cycle was found in the 'build-nudges-ref' dependency graph
```

controllers/webhooks/component_webhook.go
```diff
@@ -118,3 +141,33 @@ func (r *ComponentWebhook) ValidateDelete(ctx context.Context, obj runtime.Objec
...
+			if !k8sErrors.IsNotFound(err) {
+				return err
+			}
+		}
+
+		if nudgedComponent.Spec.Application != componentApp {
+			return fmt.Errorf("component %s cannot be added to spec.build-nudges-ref as it belongs to a different application", nudgedComponentName)
+		}
+
+		err = r.validateBuildNudgesRefGraph(ctx, nudgedComponent.Spec.BuildNudgesRef, componentNamespace, componentName, componentApp)
```
yangcao77, line 166: in this case, it only checks the component being created. so I guess we assumes the created components have no ref cycles?  
That makes sense actually.
just thinking any other edge cases can bypass the check and result in any cycles?

johnmcollier: > in this case, it only checks the component being created. so I guess we assumes the created components have no ref cycles?

Yes, we don't validate other created components, as the webhook should have already validated those components when they were first created. It saves CPU cycles on a webhook that needs to complete before the resource can be created

yangcao77 (approved):
change looks good

PR #414, opened 2023-11-09
Update codecov ignore list - ignore token mock and contract test files

johnmcollier:
Those files are test only, and shouldn't be included in coverage calculations

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -9,3 +9,4 @@ ignore:
   - "cdq-analysis/main.go" # entry point for cdq-analysis docker image
   - "pkg/github/mock.go" # mock file for testing
   - "pkg/github/token_moke.go" # mock file for testing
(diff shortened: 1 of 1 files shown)


codecov.yml
```diff
@@ -8,4 +8,5 @@ ignore:
   - "cdq-analysis/pkg/detect_mock.go" # mock file for testing
   - "cdq-analysis/main.go" # entry point for cdq-analysis docker image
   - "pkg/github/mock.go" # mock file for testing
-  - "pkg/github/token_moke.go" # mock file for testing
\ No newline at end of file
+  - "pkg/github/token_moke.go" # mock file for testing
```
maysunfaisal, line 11: token_mock.go? 😆

johnmcollier: Heh, good find.

PR #421, opened 2023-11-29
[DEVHAS-491] Mock SPI Client & Devfile Utils Client for Pvt Comp Controller Tests

maysunfaisal:
### What does this PR do?:

- Mock SPI Client to handle private Component controller tests
- Mock Devfile Util Client to hanle private Component controller tests
- Adds more use cases for private Component controller tests

https://issues.redhat.com/browse/DEVHAS-491

### PR acceptance criteria:

- [x] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact

  

Tests should pass


maysunfaisal:
Once https://github.com/devfile/library/pull/191 is merged, I should be pull in the latest commit and teh tests should pass.

controllers/component_controller.go
```diff
@@ -401,7 +403,7 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		if devfileLocation != "" {
 			// Parse the Component Devfile
 			log.Info(fmt.Sprintf("Parsing Devfile from the Devfile location %s... %v", devfileLocation, req.NamespacedName))
-			compDevfileData, err = cdqanalysis.ParseDevfileWithParserArgs(&devfileParser.ParserArgs{URL: devfileLocation, Token: gitToken})
+			compDevfileData, err = cdqanalysis.ParseDevfileWithParserArgs(&devfileParser.ParserArgs{URL: devfileLocation, Token: gitToken, DevfileUtilsClient: r.DevfileUtilsClient})
```
yangcao77, line 406: I saw in component controller, it calls multiple devfile parser. why only pass in `DevfileUtilsClient` in this place? because the test case only covers here?
also should this be tested in CDQ module as well?

controllers/component_controller_test.go
```diff
@@ -2093,6 +2094,325 @@ var _ = Describe("Component controller", func() {
...
+					Namespace: HASAppNamespace,
+				},
+				Spec: appstudiov1alpha1.ComponentSpec{
+					ComponentName: ComponentName,
+					Application:   applicationName,
+					Secret:        componentName,
+					Source: appstudiov1alpha1.ComponentSource{
+						ComponentSourceUnion: appstudiov1alpha1.ComponentSourceUnion{
+							GitSource: &appstudiov1alpha1.GitSource{
+								URL: SampleRepoLink,
```
yangcao77, line 2138: the test says for private repo, is this `samplerepolink` for public repo testing tho?

controllers/component_controller.go
maysunfaisal: Good Question. Because:
1. This is the only place in component controller which uses URL `devfileParser.ParserArgs{URL: devfileLocation ...`, other invocations use `&devfileParser.ParserArgs{Data: []byte(component.Status.Devfile) ....`
2. https://issues.redhat.com/browse/DEVHAS-492 deals with testing private links in the devfile, so when a Devfile is from a data src and has private links for URI for example, then we can provide `DevfileUtilsClient` in these calls
3. Also, devfile/library does a default assignment if its nil, https://github.com/devfile/library/blob/main/pkg/devfile/parser/parse.go#L173-L175. So even if we dont provide argument with the call here, it will always use the real implementation. When we are working on https://issues.redhat.com/browse/DEVHAS-492, we can change it to allow for mock tests.

controllers/component_controller_test.go
maysunfaisal: Actually it doesn't matter what repo we provide here, we are testing the token "valid-token" which returns a mock devfile from  See https://github.com/devfile/library/blob/main/pkg/util/mock.go#L250. But I changed this line to a private repo to avoid confusion and added a comment to the mock file. 👍

yangcao77 (approved):
looks good to me

PR #424, opened 2023-11-29
RHTAP-1134: Remove the check for valid SCM vendor

gbenhaim:
- Required for supporting components with source in Gitlab.
- The current check was over simplified and assumed the structure of the SCM's hostname, which can be anything (we can only require for it to be a url).


johnmcollier (approved):
/lgtm
/approve

PR #427, opened 2023-12-05
Rework replica logic

kim-tsao:
### What does this PR do?:

- Updates the logic so deployment replicas are set to 1 by default if implied (i.e. unset in the component spec, devfile and inline deployment).  Order of precedence is comp.spec.replicas > devfile replicas > deployment replicas
- The component attribute `deployment/replicas: 1` will be written for the default case and will trigger an update to the component model.  
- If the replicas are only specified in the deployment, no component attribute will be generated (it'll be removed if previously set).  We will rely on the value that's already specified (by parsing) rather than the one set in the attribute  
- Fixed some bugs
- Updated test cases so there's a greater coverage.  Mixed up some numbers so we're not always testing with replica 1

https://issues.redhat.com/browse/RHTAPBUGS-12

### PR acceptance criteria:

- [ ] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact

  

1.  Built the HAS image and deployed on a local cluster.  Created a token with read and write permissions to write into a test org
2. Deployed the sample [Application](https://github.com/redhat-appstudio/application-service/blob/main/config/samples/application/app.yaml) and [Component](https://github.com/redhat-appstudio/application-service/blob/main/config/samples/component/component-basic.yaml).  Verified the [replica was set to 1](https://github.com/has-test-org/application-sample-N6juw-last-relax/commit/1813a6f244b7f8af739ee082320cfe1d276e2bab) in the generated deployment
3. Updated the Component CR to set replicas to 0.  Verified the [replica was set to 0](https://github.com/has-test-org/application-sample-N6juw-last-relax/commit/fa7406d220a967aae20c968e1c1e2945d0a977c9) in the generated deployment
4. Updated the Component CR to remove the replica field.  Verified the [replica was set to 1 (value from the deployment)](https://github.com/has-test-org/application-sample-N6juw-last-relax/commit/eea0bc9198f3355dcda3935f220ee6724ec58324)
5. Created a new Component CR with the QE test and verified the [replica was 1](https://github.com/has-test-org/application-sample-N6juw-last-relax/blob/main/components/component-sample-2/base/deployment.yaml#L13) (and not 0 as it was set in the failing runs)

### DIFF
--- a/controllers/update.go
+++ b/controllers/update.go
@@ -70,24 +70,75 @@ func (r *ComponentReconciler) updateComponentDevfileModel(req ctrl.Request, hasC
 	for _, kubernetesComponent := range kubernetesComponents {
 		compUpdateRequired := false
 		// Update for Replica
-		currentReplica := 0
+		currentReplica := 1 // default value is 1
+		keyFound := true
+
 		if len(kubernetesComponent.Attributes) == 0 {
 			kubernetesComponent.Attributes = attributes.Attributes{}
+			keyFound = false
 		} else {
 			var err error
 			currentReplica = int(kubernetesComponent.Attributes.GetNumber(devfile.ReplicaKey, &err))
 			if err != nil {
 				if _, ok := err.(*attributes.KeyNotFoundError); !ok {
 					return err
+				} else {
+					keyFound = false
+					currentReplica = 1 //if an error is raised, it'll set currentReplica to 0 so we need to reset back to the default
 				}
 			}
 		}
 
-		numReplicas := util.GetIntValue(component.Spec.Replicas)
-		if currentReplica != numReplicas {
-			log.Info(fmt.Sprintf("setting devfile component %s attribute component.Spec.Replicas to %v", kubernetesComponent.Name, numReplicas))
-			kubernetesComponent.Attributes = kubernetesComponent.Attributes.PutInteger(devfile.ReplicaKey, numReplicas)
-			compUpdateRequired = true
+		numReplicas := 1 //default value
+		if component.Spec.Replicas != nil {
+			numReplicas = util.GetIntValue(component.Spec.Replicas)
+			// Component.Spec.Replicas will override any other settings.
+			// We will write the attribute if it doesn't exist for the initial creation case when comp.spec.replica is 1
+			if currentReplica != numReplicas || !keyFound {
+				log.Info(fmt.Sprintf("setting devfile component %s attribute component.Spec.Replicas to %v", kubernetesComponent.Name, numReplicas))
+				kubernetesComponent.Attributes = kubernetesComponent.Attributes.PutInteger(devfile.ReplicaKey, numReplicas)
+				compUpdateRequired = true
+			}
+		} else {
+			//check to see if we have an inlined deployment
+			isDeployReplicaSet := false
+			inlined := kubernetesComponent.Kubernetes.Inlined
+			if inlined != "" {
+				log.Info(fmt.Sprintf("reading the kubernetes inline from component %s", component.Name))
+				src := parser.YamlSrc{
+					Data: []byte(inlined),
+				}
+
+				values, err := parser.ReadKubernetesYaml(src, nil, nil)
+				if err != nil {
+					return err
+				}
+
+				resources, err := parser.ParseKubernetesYaml(values)
+				if err != nil {
+					return err
+				}
+
+				if len(resources.Deployments) > 0 {
+					replica := resources.Deployments[0].Spec.Replicas
+					if replica != nil {
+						isDeployReplicaSet = true
+						//remove the deployment/replicas attribute which can be left behind if we go from a set value to an unset value
+						if kubernetesComponent.Attributes.Exists(devfile.ReplicaKey) {
+							num := int(kubernetesComponent.Attributes.GetNumber(devfile.ReplicaKey, nil))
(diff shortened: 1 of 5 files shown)


controllers/update.go
```diff
@@ -70,24 +70,75 @@ func (r *ComponentReconciler) updateComponentDevfileModel(req ctrl.Request, hasC
...
-			log.Info(fmt.Sprintf("setting devfile component %s attribute component.Spec.Replicas to %v", kubernetesComponent.Name, numReplicas))
-			kubernetesComponent.Attributes = kubernetesComponent.Attributes.PutInteger(devfile.ReplicaKey, numReplicas)
-			compUpdateRequired = true
+		numReplicas := 1 //default value
+		if component.Spec.Replicas != nil {
+			numReplicas = util.GetIntValue(component.Spec.Replicas)
+			// Component.Spec.Replicas will override any other settings.
+			// We will write the attribute if it doesn't exist for the initial creation case when comp.spec.replica is 1
+			if currentReplica != numReplicas || !keyFound {
+				log.Info(fmt.Sprintf("setting devfile component %s attribute component.Spec.Replicas to %v", kubernetesComponent.Name, numReplicas))
```
maysunfaisal, line 98:
```suggestion
				log.Info(fmt.Sprintf("setting devfile component %s attribute <devfile.ReplicaKey>; to %v", kubernetesComponent.Name, numReplicas))
```
whatever the key is perhaps

controllers/update.go
```diff
@@ -70,24 +70,75 @@ func (r *ComponentReconciler) updateComponentDevfileModel(req ctrl.Request, hasC
...
+					return err
+				}
+
+				if len(resources.Deployments) > 0 {
+					replica := resources.Deployments[0].Spec.Replicas
+					if replica != nil {
+						isDeployReplicaSet = true
+						//remove the deployment/replicas attribute which can be left behind if we go from a set value to an unset value
+						if kubernetesComponent.Attributes.Exists(devfile.ReplicaKey) {
+							num := int(kubernetesComponent.Attributes.GetNumber(devfile.ReplicaKey, nil))
```
maysunfaisal, line 128:
```suggestion
							num := int(kubernetesComponent.Attributes.GetNumber(devfile.ReplicaKey, &err))
```
I dont think we should ignore the err

pkg/devfile/devfile.go
```diff
@@ -124,14 +124,27 @@ func GetResourceFromDevfile(log logr.Logger, devfileData data.DevfileData, deplo
...
+						} else {
+							// check the deployment to see what value is set
+							replica := resources.Deployments[0].Spec.Replicas
+							if replica != nil {
+								// if there is no component attribute, we will use the value in the deployment spec
+								currentReplica = *replica
+							} else {
+								// default value is 1.  We shouldn't hit this code path since this check is done in the update logic and will write the attribute deployment/replicas:1
+								currentReplica = 1
+							}
```
maysunfaisal, line 142: I think these lines can be simplified?
1. we are checking if replica is set in deploy
2. if it is, we are reading it and setting it back 🤔 

Even when we are checking if its empty and setting it to 1, do we even need to? Its 1 by default in Kubernetes i.e. if you were to create a pod without this property there is only one repllca, no?

controllers/update.go
```diff
@@ -70,24 +70,75 @@ func (r *ComponentReconciler) updateComponentDevfileModel(req ctrl.Request, hasC
...
+			keyFound = false
 		} else {
 			var err error
 			currentReplica = int(kubernetesComponent.Attributes.GetNumber(devfile.ReplicaKey, &err))
 			if err != nil {
 				if _, ok := err.(*attributes.KeyNotFoundError); !ok {
 					return err
+				} else {
+					keyFound = false
+					currentReplica = 1 //if an error is raised, it'll set currentReplica to 0 so we need to reset back to the default
```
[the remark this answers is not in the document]
kim-tsao: I added this comment due to test observations and what's documented for the `Getnumber` function.  In the case of any error (KeyNotFoundError included), it'll assign 0 which is why we need to reset

controllers/update.go
kim-tsao: I assumed no errors at this point if key is found.  I guess there may be other errors

pkg/devfile/devfile.go
kim-tsao: It looks redundant but it's because we are searching for the component attribute key first in line 129 

`currentReplica := int32(component.Attributes.GetNumber(ReplicaKey, &replicaErr))`

We are relying on the presence of the attribute to override anything that's set in deployment but if it doesn't exist, it'll return a `KeyNotFoundError` and assign a 0 value.  Similar to the previous comment it's there to "reset" the value

controllers/update.go
```diff
@@ -70,24 +70,75 @@ func (r *ComponentReconciler) updateComponentDevfileModel(req ctrl.Request, hasC
 	for _, kubernetesComponent := range kubernetesComponents {
 		compUpdateRequired := false
 		// Update for Replica
-		currentReplica := 0
+		currentReplica := 1 // default value is 1
```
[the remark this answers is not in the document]
kim-tsao: It is for the initial value but it gets assigned to the component attribute value later which is the "current" value

maysunfaisal (approved):

PR #433, opened 2024-01-09
Filter out user devfile errors

maysunfaisal:
### What does this PR do?:

- Filters out the user errors in a devfile so that the prometheus metrics reflect a proper state
- Adds new test case to test the prometheus metric

https://issues.redhat.com/browse/DEVHAS-580

### PR acceptance criteria:

- [x] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact

  

Test with updated metrics should pass

### DIFF
--- a/controllers/applicationsnapshotenvironmentbinding_controller_test.go
+++ b/controllers/applicationsnapshotenvironmentbinding_controller_test.go
@@ -2371,7 +2371,7 @@ var _ = Describe("SnapshotEnvironmentBinding controller", func() {
 				return len(createdBinding.Status.GitOpsRepoConditions) > 0 && createdBinding.Status.GitOpsRepoConditions[0].Reason == "GenerateError"
 			}, timeout, interval).Should(BeTrue())
 
-			Expect(createdBinding.Status.GitOpsRepoConditions[0].Message).Should(ContainSubstring("failed to decode devfile json"))
+			Expect(createdBinding.Status.GitOpsRepoConditions[0].Message).Should(ContainSubstring("cannot unmarshal string into Go value of type map[string]interface"))
 
 			// Delete the specified HASComp resource
 			deleteHASCompCR(hasCompLookupKey)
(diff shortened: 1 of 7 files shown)


controllers/applicationsnapshotenvironmentbinding_controller_test.go
```diff
@@ -2371,7 +2371,7 @@ var _ = Describe("SnapshotEnvironmentBinding controller", func() {
 				return len(createdBinding.Status.GitOpsRepoConditions) > 0 && createdBinding.Status.GitOpsRepoConditions[0].Reason == "GenerateError"
 			}, timeout, interval).Should(BeTrue())
 
-			Expect(createdBinding.Status.GitOpsRepoConditions[0].Message).Should(ContainSubstring("failed to decode devfile json"))
+			Expect(createdBinding.Status.GitOpsRepoConditions[0].Message).Should(ContainSubstring("cannot unmarshal string into Go value of type map[string]interface"))
```
johnmcollier, line 2374: What does the full error message look like with this? The test just checks for the substring, so I wasn't sure.

If it's just `cannot unmarshal string into Go value of type map[string]interface`, it's not really user friendly (compared to the old message), and we should probably update the error string that's set in the SEB status. If it's just a subset, and it mentions something along the lines of "failed to decode devfile" or "failed to parse devfile", then that's fine.

controllers/component_controller_test.go
```diff
@@ -792,7 +792,7 @@ var _ = Describe("Component controller", func() {
 
 			errCondition := updatedHasComp.Status.Conditions[len(updatedHasComp.Status.Conditions)-1]
 			Expect(errCondition.Status).Should(Equal(metav1.ConditionFalse))
-			Expect(errCondition.Message).Should(ContainSubstring("failed to decode devfile json"))
+			Expect(errCondition.Message).Should(ContainSubstring("cannot unmarshal string into Go value of type map[string]interface"))
```
johnmcollier, line 795: Same comment that I left for SEB

controllers/applicationsnapshotenvironmentbinding_controller_test.go
maysunfaisal: it looks like,

```
Expected
        <string>: Component updated failed: error parsing devfile because of non-compliant data due to json: cannot unmarshal string into Go value of type map[string]interface {}
    to contain substring
        <string>: cannot unmarslhal string into Go value of type map[string]interface
```

WDYT?

johnmcollier: Looks good

johnmcollier (approved):

PR #450, opened 2024-03-12
RHTAPSRE-467: Support the configuration of provider URL

gbenhaim:
When onboarding self-hosted SCM PAC can't determine the the base url, so it should be provided explicitly.
Allow users to mention the provider (SCM) URL using an annotation on the component.

### DIFF
--- a/gitops/generate_build.go
+++ b/gitops/generate_build.go
@@ -40,6 +40,7 @@ const (
 
 	PaCAnnotation                     = "pipelinesascode"
 	GitProviderAnnotationName         = "git-provider"
+	GitProviderAnnotationURL          = "git-provider-url"
 	PipelinesAsCodeWebhooksSecretName = "pipelines-as-code-webhooks-secret"
 	PipelinesAsCode_githubAppIdKey    = "github-application-id"
 	PipelinesAsCode_githubPrivateKey  = "github-private-key"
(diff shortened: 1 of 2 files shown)


johnmcollier (approved):
Changes look fine to me. Though I will wait for @mmorhun to review before merging.

I wasn't aware this code was still in HAS, but looks like it's called from the build-service.

@mmorhun we should seriously consider moving this code to the build-service. It's only dependency within HAS is [prepare.go](https://github.com/redhat-appstudio/application-service/blob/main/gitops/prepare/prepare.go), which AFAICT, is not used within HAS at all.

gbenhaim:
thanks for the review @johnmcollier. I fixed the formatting.

gbenhaim:
@johnmcollier I didn't understand from the test logs if it failed because of my change or because of an infra issue/flaky test. Can you please help me debug it?

johnmcollier:
@gbenhaim The e2e test failure seems to have been a flake that's resolved itself. I see some flakiness in our controller tests as well, so I've retested those to confirm

mmorhun:
@gbenhaim why cannot we retrieve git provider URL from the Component source URL?

gbenhaim:
@mmorhun I followed the configuration fields PAC provides. Since PAC doesn't retrieve the SCM host for self-hosted SCMs from the repository url, I don't think we should do it as well.

Nothing promises that the hostname of the SCM API server can be computed from the repository url, for example in a case where cloning is done through a proxy.

gitops/generate_build.go
```diff
@@ -89,25 +90,26 @@ func GeneratePACRepository(component appstudiov1alpha1.Component, config map[str
 
 	isAppUsed := IsPaCApplicationConfigured(gitProvider, config)
 
-	var gitProviderConfig *pacv1alpha1.GitProvider = nil
+	var gitProviderConfig *pacv1alpha1.GitProvider = &pacv1alpha1.GitProvider{}
```
mmorhun, line 93: Are you sure that passing empty config instead of `nil` won't change PaC behavior?

mmorhun (approved):
LGTM

gitops/generate_build.go
gbenhaim: I don't think it would make any difference but to be on the safe side I updated the PR and kept the default `gitProviderConfig` to be `nil`.

PR #451, opened 2024-03-13
Avoid race coditions from component deletion for applications already marked to be deleted

thepetk:
### What does this PR do?:

This PR adds an extra condition upon component deletion. More detailed in https://github.com/redhat-appstudio/application-service/blob/main/controllers/component_controller.go#L191 we now check if the component's application has a deletion timestamp set. If yes we don't delete the component to avoid race conditions between the component & the application deletion.

https://issues.redhat.com/browse/DEVHAS-627

### PR acceptance criteria:

- [ ] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact

### DIFF
--- a/controllers/component_controller.go
+++ b/controllers/component_controller.go
@@ -188,8 +188,8 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 			log.Info(fmt.Sprintf("added the finalizer %v", req.NamespacedName))
 		}
 	} else {
-		if hasApplication.Status.Devfile != "" && (forceGenerateGitopsResource || len(component.Status.Conditions) > 0 && component.Status.Conditions[len(component.Status.Conditions)-1].Status == metav1.ConditionTrue && containsString(component.GetFinalizers(), compFinalizerName)) {
-			// only attempt to finalize and update the gitops repo if an Application is present & the previous Component status is good
+		if hasApplication.Status.Devfile != "" && hasApplication.ObjectMeta.DeletionTimestamp.IsZero() && (forceGenerateGitopsResource || len(component.Status.Conditions) > 0 && component.Status.Conditions[len(component.Status.Conditions)-1].Status == metav1.ConditionTrue && containsString(component.GetFinalizers(), compFinalizerName)) {
+			// only attempt to finalize and update the gitops repo if an Application is present & not under deletion & the previous Component status is good
 			// A finalizer is present for the Component CR, so make sure we do the necessary cleanup steps
 			metrics.ComponentDeletionTotalReqs.Inc()
 			if err := r.Finalize(ctx, &component, &hasApplication, ghClient, gitToken); err != nil {
--- a/controllers/component_controller.go
+++ b/controllers/component_controller.go
@@ -188,14 +188,9 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 			log.Info(fmt.Sprintf("added the finalizer %v", req.NamespacedName))
 		}
 	} else {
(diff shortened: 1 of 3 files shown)


thepetk:
WIP until is tested on a local cluster

johnmcollier (approved):
Changes do look fine, but let's hold off on merging while the WIP is still present 🙂 
/lgtm

maysunfaisal (commented):
no tests needed?

thepetk:
> no tests needed?

Yeap is WIP as I'm still working on the test cases.

cc @maysunfaisal @johnmcollier

thepetk:
@johnmcollier @maysunfaisal I've removed the WIP label and added a test case for component deletion for applications that have been marked as deleted

controllers/component_controller.go
```diff
@@ -189,8 +189,13 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		}
 	} else {
 		if hasApplication.Status.Devfile != "" && (forceGenerateGitopsResource || len(component.Status.Conditions) > 0 && component.Status.Conditions[len(component.Status.Conditions)-1].Status == metav1.ConditionTrue && containsString(component.GetFinalizers(), compFinalizerName)) {
-			// only attempt to finalize and update the gitops repo if an Application is present & the previous Component status is good
+			// only attempt to finalize and update the gitops repo if an Application is present & not under deletion & the previous Component status is good
 			// A finalizer is present for the Component CR, so make sure we do the necessary cleanup steps
+			if !hasApplication.ObjectMeta.DeletionTimestamp.IsZero() {
```
maysunfaisal, line 194: LGTM but honestly I am wondering what happens when Application Deletion Timestamp is 0 and we proceed but by the time the Component Deletion is happening; Application Deleteion Timestamp is now not 0 and is faster than a Component Delete?

maysunfaisal: I think to be completely error proof, we may also need to block out Application Delete if Component Deletion is under process... 🤔

WDYT?

thepetk: It makes sense. A question I have is what if both `application` and `component` have `deletiontimestamp` not 0 and we ignore the deletion for both `application_controller` and `component_controller`. Will we keep retrying to delete `component` and/or `application`?

I was thinking if it makes sense to proceed only with the fix for the component deletion failure alert and if the issue still presists then to apply the same fix for the application too. WDYT?

maysunfaisal: If both have deletiontimestamps not 0, then we only proceed with the Application...

If you think its going to be take more time to fix it, we can check this in.

/lgtm

controllers/component_controller.go
```diff
@@ -189,8 +189,13 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+				log.Info("application %v is under deletion. Skipping deletion for component %v", hasApplication.Name, component.Name)
+				return ctrl.Result{}, nil
```
johnmcollier, line 196: Just had a thought.

If we return here, we never remove the finalizer off the component resource farther below:

https://github.com/redhat-appstudio/application-service/blob/52908910fc379bf5731a952c2f2093b886f93eb7/controllers/component_controller.go#L217

If that never happens, the Component CR can't be deleted, as the finalizer blocks deletion. We still need to remove the finalizer in this case.

johnmcollier (changes requested):

controllers/component_controller.go
thepetk: @johnmcollier I've updated the code in order to ensure that the finalizer is removed & we ignore the deletion for this case

maysunfaisal (approved):

maysunfaisal (commented):
/lgtm

PR #453, opened 2024-03-18
Add availability metrics

maysunfaisal:
### What does this PR do?:

- Adds in the Availability metrics as per https://github.com/redhat-appstudio/o11y/tree/main/exporters/dsexporter 
- Inspired from https://github.com/konflux-ci/remote-secret/pull/338 
- Metric format specification at https://docs.google.com/document/d/193FO8DYP8zCzzued53DiIT84xmyTm51v_OqiWzQi8Bg/edit
- Makes a simple noop Git API call every 20 minutes (see screenshot below)

https://issues.redhat.com/browse/DEVHAS-597

### PR acceptance criteria:

- [x] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact

  

The HAS pod will print the statements about availability:

Screenshot 2024-03-18 at 4 17 01 PM

### DIFF
--- a/pkg/availability/git.go
+++ b/pkg/availability/git.go
@@ -1,0 +1,71 @@
+//
+// Copyright 2024 Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package availability
+
+import (
+	"context"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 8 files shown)


maysunfaisal:
Nit: Should not check availability every second, need to update it to a more broad duration like every hour. Otherwise, the Git token will be rate limited!

thepetk:
> Nit: Should not check availability every second, need to update it to a more broad duration like every hour. Otherwise, the Git token will be rate limited!

+1 I think every second is too much right? I would expect something like 1 to 5 minutes. As github's API is a service that we can rely on I'd suggest 5 to 10 mins.

pkg/availability/git.go
```diff
@@ -0,0 +1,71 @@
...
+				return
+			}
+		}
+	}()
+	return nil
+}
+
+func (r *AvailabilityWatchdog) checkAvailability(ctx context.Context) {
+	log := ctrl.LoggerFrom(ctx)
+	checkGitLabel := prometheus.Labels{"check": "github"}
```
johnmcollier, line 55: Do we need to update the metric specification doc? It HAS
> has_availability{other_label=”sso”} | konflux_up(service=”has”, check=”sso”)

But `check=sso` doesn't make much sense, and what you have here (`github`), makes more sense

maysunfaisal: updated the doc

johnmcollier (approved):

PR #454, opened 2024-03-19
Return error for non supported development platform urls

thepetk:
### What does this PR do?:

The PR updates the `component_controller` to return a user error in case the `given` URL is not from a supported development platform (right now only github).

The necessary test cases have been added.

https://issues.redhat.com/browse/DEVHAS-628

### PR acceptance criteria:

- [ ] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact

### DIFF
--- a/cdq-analysis/pkg/util.go
+++ b/cdq-analysis/pkg/util.go
@@ -170,11 +170,8 @@ func ConvertGitHubURL(URL string, revision string, context string) (string, erro
 	if err != nil {
 		return "", &InvalidURL{URL: URL, Err: err}
 	}
-	if !strings.Contains(url.Host, "github") {
-		return "", &InvalidURL{URL: URL, Err: fmt.Errorf("url %v is not from github.com", URL)}
-	}
 
-	if !strings.Contains(url.Host, "raw") {
+	if strings.Contains(url.Host, "github") && !strings.Contains(url.Host, "raw") {
 		// Convert path part of the URL
 		URLSlice := strings.Split(URL, "/")
 		if len(URLSlice) > 2 && URLSlice[len(URLSlice)-2] == "tree" {
--- a/cdq-analysis/pkg/util.go
+++ b/cdq-analysis/pkg/util.go
@@ -171,7 +171,7 @@ func ConvertGitHubURL(URL string, revision string, context string) (string, erro
 		return "", &InvalidURL{URL: URL, Err: err}
 	}
 	if !strings.Contains(url.Host, "github") {
-		return "", &InvalidURL{URL: URL, Err: fmt.Errorf("url %v not from github.com")}
+		return "", &InvalidURL{URL: URL, Err: fmt.Errorf("url %v is not from github.com", URL)}
 	}
 
 	if !strings.Contains(url.Host, "raw") {
--- a/cdq-analysis/pkg/util.go
+++ b/cdq-analysis/pkg/util.go
@@ -170,8 +170,11 @@ func ConvertGitHubURL(URL string, revision string, context string) (string, erro
 	if err != nil {
 		return "", &InvalidURL{URL: URL, Err: err}
 	}
+	if !strings.Contains(url.Host, "github") {
+		return "", &InvalidURL{URL: URL, Err: fmt.Errorf("url %v not from github.com")}
+	}
 
-	if strings.Contains(url.Host, "github") && !strings.Contains(url.Host, "raw") {
+	if !strings.Contains(url.Host, "raw") {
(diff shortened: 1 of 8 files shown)


cdq-analysis/pkg/util.go
```diff
@@ -170,8 +170,11 @@ func ConvertGitHubURL(URL string, revision string, context string) (string, erro
 	if err != nil {
 		return "", &InvalidURL{URL: URL, Err: err}
 	}
+	if !strings.Contains(url.Host, "github") {
```
maysunfaisal, line 173: QQ: how did you test this change? 

This is fine with CLI/test suite import cases. Does it also work with Konflux UI? Because that uses CDQ first before going to Component controller... 🤔

thepetk: Oh ok. TBH I haven't changed it with the Konflux UI. Having in mind what you mention, I think is better to include this change only on the component controller side right?

thepetk: @maysunfaisal following your suggestion I've moved the functionality to the `component_controller` side.

controllers/component_controller.go
```diff
@@ -298,6 +299,13 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		var devfileBytes []byte
 
 		if source.GitSource != nil && source.GitSource.URL != "" {
+			if err := validateGithubURL(source.GitSource.URL); err != nil {
+				// User error - the git url provided is not from github
+				log.Error(err, "unable to validate github url")
+				metrics.IncrementComponentCreationSucceeded("", "")
+				_ = r.SetCreateConditionAndUpdateCR(ctx, req, &component, err)
+				return ctrl.Result{}, err
```
maysunfaisal, line 307:
```suggestion
				return ctrl.Result{}, nil
```

maysunfaisal: we should not requeue

controllers/component_controller.go
```diff
@@ -884,3 +892,17 @@ func checkForCreateReconcile(component appstudiov1alpha1.Component) (bool, strin
 	// If there are no Conditions or no Updated Condition, then this is a Create reconcile
 	return true, errCondition
 }
+
+// isGithubURL checks if the given url includes github in hostname
+// In case of invalid url (not able to parse) returns false.
+func validateGithubURL(URL string) error {
```
maysunfaisal, line 898: would probably move it to pkg/util

maysunfaisal (commented):
So does Gitlab url get blocked from the UI?

maysunfaisal:
> > So does Gitlab url get blocked from the UI?
> 
> @maysunfaisal I've updated the PR with your suggestions. Regarding the gitlab provider from the `rhtap` side I see
> 
> Screenshot from 2024-03-20 16-16-03

1. I tried this repo, it passes via UI because there is no devfile and we match a devfile from devfile registry... When this happens, only git cli is used which works for both github and gitlab but the build never happens, i see this: 
Screenshot 2024-03-20 at 3 31 24 PM

2. If you take a gitlab repo which has a devfile https://gitlab.com/maysun/devfilesample, CDQ will fail: 
Screenshot 2024-03-20 at 3 24 37 PM

and you will need to choose a Runtime from dropdown which will revert to point 1 where devfile is matched from devfile registry

---

so, we will need to also block CDQ controller from accepting Gitlab URLs like how you did for Component controller..

thepetk:
> > > So does Gitlab url get blocked from the UI?
> > 
> > 
> > @maysunfaisal I've updated the PR with your suggestions. Regarding the gitlab provider from the `rhtap` side I see
> > Screenshot from 2024-03-20 16-16-03
> 
> 1. I tried this repo, it passes via UI because there is no devfile and we match a devfile from devfile registry... When this happens, only git cli is used which works for both github and gitlab but the build never happens, i see this:
> 
> Screenshot 2024-03-20 at 3 31 24 PM
> 2. If you take a gitlab repo which has a devfile https://gitlab.com/maysun/devfilesample, CDQ will fail:
> 
> Screenshot 2024-03-20 at 3 24 37 PM
> and you will need to choose a Runtime from dropdown which will revert to point 1 where devfile is matched from devfile registry
> 
> so, we will need to also block CDQ controller from accepting Gitlab URLs like how you did for Component controller..

@maysunfaisal I've pushed similar logic to the `componentdetectionquery_controller.go` and added new test cases to `componentdetectionquery_controller_test.go`

maysunfaisal (approved):
/lgtm

PR #459, opened 2024-03-27
Fix metrics for component creation and gitlab url

thepetk:
### What does this PR do?:

This PR tries to fix the metrics for component creation for cases of gitlab URLs by returning a user error on the `component_controller.go`

https://issues.redhat.com/browse/DEVHAS-628

### PR acceptance criteria:

- [ ] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact

### DIFF
--- a/controllers/component_controller.go
+++ b/controllers/component_controller.go
@@ -295,6 +295,7 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	if component.Status.Devfile == "" {
 
 		source := component.Spec.Source
+		gitSourceFromGitlab := false
 
 		var compDevfileData data.DevfileData
 		var devfileLocation string
@@ -304,9 +305,7 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 			if err := util.ValidateGithubURL(source.GitSource.URL); err != nil {
 				// User error - the git url provided is not from github
 				log.Error(err, "unable to validate github url")
-				metrics.IncrementComponentCreationSucceeded("", "")
-				_ = r.SetCreateConditionAndUpdateCR(ctx, req, &component, err)
-				return ctrl.Result{}, nil
+				gitSourceFromGitlab = true
 			}
 			context := source.GitSource.Context
 			// If a Git secret was passed in, retrieve it for use in our Git operations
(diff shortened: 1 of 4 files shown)


controllers/component_controller.go
```diff
@@ -301,6 +301,13 @@ func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		var devfileBytes []byte
 
 		if source.GitSource != nil && source.GitSource.URL != "" {
+			if err := util.ValidateGithubURL(source.GitSource.URL); err != nil {
+				// User error - the git url provided is not from github
+				log.Error(err, "unable to validate github url")
+				metrics.IncrementComponentCreationSucceeded("", "")
+				_ = r.SetCreateConditionAndUpdateCR(ctx, req, &component, err)
+				return ctrl.Result{}, nil
+			}
```
maysunfaisal, line 310: I think this will still be executed via HAC UI? Can you confirm by just building the HAS image and creating a Component via CLI?

We should not block gitlab errors out like returning over here, ref https://redhat-internal.slack.com/archives/C02HZRGKDEY/p1711484683757149?thread_ts=1711484510.936249&cid=C02HZRGKDEY

I think we should have a flag or something which indicates its a Gitlab repo and anytime there is an error with accessing the Gitlab repo because we are using Github API, then we can ignore that in our metrics count and return as a normal user error..

@johnmcollier thoughts?

johnmcollier: Yeah, I'd avoid returning here, and just let the controller handle the URL as it was before.

johnmcollier (changes requested):
Agreed with Maysun's concern.

What about just not incrementing any metrics for components that have a GitLab repo in their source? We mgiht be losing some metrics, but given that HAS doesn't support GitLab components, I don't think we'd be losing much?

thepetk:
> Agreed with Maysun's concern.
> 
> What about just not incrementing any metrics for components that have a GitLab repo in their source? We might be losing some metrics, but given that HAS doesn't support GitLab components, I don't think we'd be losing much?

Yes I think if we don't want to return an error in case of a gitlab URL at all it makes sense only to take care of the metrics. Will try to update the PR early next week. Thanks for the help!

controllers/component_controller.go
thepetk: @johnmcollier & @maysunfaisal I've tried to remove the early return and add the `gitSourceFromGitlab`.

Tested locally to create a `component` with gitlab URL and had got `"Component create failed: unable to get default branch of Github Repo"` error which is the "normal" error we got in the past. The difference here is that if the `gitSourceFromGitlab` is `true` we increase the successful component creation metric instead of the component creation failed one.

I've updated the test cases to check the metrics count is ok.

maysunfaisal (approved):
/lgtm

I think there are few other places we could handle for Gitlab URLs but at this stage of the development cycle, I am inclined to not touch more code base for something we dont support. This change looks like it will help alleviate metric fallures for HAS (with CDQ).

johnmcollier (approved):

PR #460, opened 2024-04-01
Avoid application deletion race conditions

thepetk:
### What does this PR do?:
This PR tries to implement the same logic followed to the `component_controller` (introduced here: https://github.com/redhat-appstudio/application-service/pull/451) on the `application_controller` too. This way we avoid race conditions upon `application` & `component` deletion.

More detailed, we are ensuring that before the `application` deletion (after we have confirmed the presence of a finalizer), that the application has no components under deletion. If it does we are requeueing the deletion.

https://issues.redhat.com/browse/DEVHAS-627

### PR acceptance criteria:

- [ ] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact

### DIFF
--- a/controllers/application_controller.go
+++ b/controllers/application_controller.go
@@ -117,7 +117,7 @@ func (r *ApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 			var components appstudiov1alpha1.ComponentList
 			err = r.List(ctx, &components)
 			if err != nil {
-				log.Error(err, "failed to check if any application component is under deletion")
+				log.Error(err, "failed to check if any Component is under deletion, proceeding with Application delete")
 			}
 			if err := util.VerifyNoApplicationComponentUnderDeletion(components, application.Name); err != nil {
 				// requeue in case any of the application components are under deletion
--- a/controllers/application_controller.go
+++ b/controllers/application_controller.go
@@ -114,6 +114,16 @@ func (r *ApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		}
 	} else {
 		if containsString(application.GetFinalizers(), appFinalizerName) {
+			var components appstudiov1alpha1.ComponentList
+			err = r.List(context.Background(), &components)
+			if err != nil {
+				log.Error(err, "failed to check if any application component is under deletion")
+			}
+			if err := util.VerifyNoApplicationComponentUnderDeletion(components, application.Name); err != nil {
+				// requeue in case any of the application components are under deletion
+				log.Error(err, "Error an application component is under deletion. Deletion requeued")
+				return reconcile.Result{}, err
+			}
 			metrics.ApplicationDeletionTotalReqs.Inc()
 			// A finalizer is present for the Application CR, so make sure we do the necessary cleanup steps
 			if finalizeErr := r.Finalize(ctx, &application, ghClient); finalizeErr != nil {
--- a/controllers/application_controller.go
+++ b/controllers/application_controller.go
@@ -121,7 +121,7 @@ func (r *ApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 			}
 			if err := util.VerifyNoApplicationComponentUnderDeletion(components, application.Name); err != nil {
 				// requeue in case any of the application components are under deletion
-				log.Error(err, "Error an application component is under deletion. Deletion requeued")
+				log.Error(err, "Component is under deletion, Application deletion requeued")
 				return reconcile.Result{}, err
 			}
 			metrics.ApplicationDeletionTotalReqs.Inc()
--- a/controllers/application_controller.go
+++ b/controllers/application_controller.go
@@ -115,7 +115,7 @@ func (r *ApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	} else {
 		if containsString(application.GetFinalizers(), appFinalizerName) {
 			var components appstudiov1alpha1.ComponentList
-			err = r.List(context.Background(), &components)
+			err = r.List(ctx, &components)
 			if err != nil {
 				log.Error(err, "failed to check if any application component is under deletion")
 			}
--- a/controllers/application_controller_test.go
+++ b/controllers/application_controller_test.go
@@ -388,9 +388,6 @@ var _ = Describe("Application controller", func() {
 				Spec: appstudiov1alpha1.ApplicationSpec{
 					DisplayName: DisplayName,
 					Description: Description,
-					AppModelRepository: appstudiov1alpha1.ApplicationGitRepository{
-						URL: "https://github.com/testorg/petclinic-app",
-					},
 				},
 			}
 
--- a/controllers/application_controller_test.go
+++ b/controllers/application_controller_test.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/devfile/library/v2/pkg/devfile/parser"
 	"github.com/prometheus/client_golang/prometheus/testutil"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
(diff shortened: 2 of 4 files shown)


controllers/application_controller.go
```diff
@@ -114,6 +114,16 @@ func (r *ApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		}
 	} else {
 		if containsString(application.GetFinalizers(), appFinalizerName) {
+			var components appstudiov1alpha1.ComponentList
+			err = r.List(context.Background(), &components)
```
maysunfaisal, line 118:
```suggestion
			err = r.List(ctx &components)
```

controllers/application_controller.go
```diff
@@ -114,6 +114,16 @@ func (r *ApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+			if err != nil {
+				log.Error(err, "failed to check if any application component is under deletion")
```
maysunfaisal, line 120:
```suggestion
				log.Error(err, "failed to check if any Component is under deletion, proceeding with Application delete")
```

controllers/application_controller.go
```diff
@@ -114,6 +114,16 @@ func (r *ApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+			}
+			if err := util.VerifyNoApplicationComponentUnderDeletion(components, application.Name); err != nil {
+				// requeue in case any of the application components are under deletion
+				log.Error(err, "Error an application component is under deletion. Deletion requeued")
```
maysunfaisal, line 124:
```suggestion
				log.Error(err, "Component is under deletion, Application deletion requeued")
```

maysunfaisal (commented):
small changes

maysunfaisal (approved):
/lgtm

johnmcollier:
> More detailed, we are ensuring that before the application deletion (after we have confirmed the presence of a finalizer), that the application has no components under deletion. If it does we are requeueing the deletion.

I'm wondering if this is needed? Since we fixed the issue where the Component deletion could get stuck if the GitOps repository has been deleted, I don't think there's a potential race condition that can occur here?

johnmcollier (commented):
> More detailed, we are ensuring that before the application deletion (after we have confirmed the presence of a finalizer), that the application has no components under deletion. If it does we are requeueing the deletion.

I'm wondering if this is needed? Since we fixed the issue where the Component deletion could get stuck if the GitOps repository has been deleted, I don't think there's a potential race condition that can occur here? If there's Components that are under deletion when we delete the Application, that just means the GitOps repository will get deleted before the Component finalizer runs (and since https://github.com/redhat-appstudio/application-service/pull/451 was merged, that shouldn't be an issue).

thepetk:
> > More detailed, we are ensuring that before the application deletion (after we have confirmed the presence of a finalizer), that the application has no components under deletion. If it does we are requeueing the deletion.
> 
> I'm wondering if this is needed? Since we fixed the issue where the Component deletion could get stuck if the GitOps repository has been deleted, I don't think there's a potential race condition that can occur here? If there's Components that are under deletion when we delete the Application, that just means the GitOps repository will get deleted before the Component finalizer runs (and since #451 was merged, that shouldn't be an issue).
> 
> We remove the finalizer no matter what [here](https://github.com/thepetk/application-service/blob/a08a1287423736befe5cebdedb70bf100616f4db/controllers/component_controller.go#L210-L222), so even if the Component clean up "fails" due to a missing GitOps repository, the deletion of either resources won't be blocked



> > More detailed, we are ensuring that before the application deletion (after we have confirmed the presence of a finalizer), that the application has no components under deletion. If it does we are requeueing the deletion.
> 
> I'm wondering if this is needed? Since we fixed the issue where the Component deletion could get stuck if the GitOps repository has been deleted, I don't think there's a potential race condition that can occur here? If there's Components that are under deletion when we delete the Application, that just means the GitOps repository will get deleted before the Component finalizer runs (and since #451 was merged, that shouldn't be an issue).
> 
> We remove the finalizer no matter what [here](https://github.com/thepetk/application-service/blob/a08a1287423736befe5cebdedb70bf100616f4db/controllers/component_controller.go#L210-L222), so even if the Component clean up "fails" due to a missing GitOps repository, the deletion of either resources won't be blocked

FYI the PR is basically the extra step discussed here: https://github.com/redhat-appstudio/application-service/pull/451#discussion_r1529559854

The main reason is that we want to fix the component deletion metrics. I've tried to revisit the investigation on the component deletion errors after your comment. My notes:

* The truth is In the latest error logs I cannot see any `Unable to update GitOps repository for component..` which comes from an unsuccessful deletion.

* As we have already noticed the main log is: `Operation cannot be fulfilled on...the object has been modified;`. I think this error comes from [line#213](https://github.com/redhat-appstudio/application-service/blob/main/controllers/component_controller.go#L213) where we are trying to update the `to-be-deleted` component.

* I can see that in [line#194](https://github.com/redhat-appstudio/application-service/blob/main/controllers/component_controller.go#L194) we are increasing the total deletion attempts. But we only increase the component deletion failed/succeeded for specific cases (see [line#205](https://github.com/redhat-appstudio/application-service/blob/main/controllers/component_controller.go#L205) & [line#219](https://github.com/redhat-appstudio/application-service/blob/main/controllers/component_controller.go#L219)).

* Having in mind the above note I think the reason of the component deletion not successful alert might be coming because our total attempts are much bigger than the sum of failed + succeeded (e.g we increase the total attempts but we don't equally increase the failed/successful). As a result the succeeded attempts should be only a small pertecantge of the total deletion attempts.

From the above notes I'd say we could:

* Increase total attempts only for cases we decide that the attempt is failed/succeded
OR
* We add succeded attempts for lines [214](https://github.com/redhat-appstudio/application-service/blob/main/controllers/component_controller.go#L214) & [216](https://github.com/redhat-appstudio/application-service/blob/main/controllers/component_controller.go#L216)

wdyt? cc @maysunfaisal @johnmcollier

thepetk:
> > > More detailed, we are ensuring that before the application deletion (after we have confirmed the presence of a finalizer), that the application has no components under deletion. If it does we are requeueing the deletion.
> > 
> > 
> > I'm wondering if this is needed? Since we fixed the issue where the Component deletion could get stuck if the GitOps repository has been deleted, I don't think there's a potential race condition that can occur here? If there's Components that are under deletion when we delete the Application, that just means the GitOps repository will get deleted before the Component finalizer runs (and since #451 was merged, that shouldn't be an issue).
> > We remove the finalizer no matter what [here](https://github.com/thepetk/application-service/blob/a08a1287423736befe5cebdedb70bf100616f4db/controllers/component_controller.go#L210-L222), so even if the Component clean up "fails" due to a missing GitOps repository, the deletion of either resources won't be blocked
> 
> > > More detailed, we are ensuring that before the application deletion (after we have confirmed the presence of a finalizer), that the application has no components under deletion. If it does we are requeueing the deletion.
> > 
> > 
> > I'm wondering if this is needed? Since we fixed the issue where the Component deletion could get stuck if the GitOps repository has been deleted, I don't think there's a potential race condition that can occur here? If there's Components that are under deletion when we delete the Application, that just means the GitOps repository will get deleted before the Component finalizer runs (and since #451 was merged, that shouldn't be an issue).
> > We remove the finalizer no matter what [here](https://github.com/thepetk/application-service/blob/a08a1287423736befe5cebdedb70bf100616f4db/controllers/component_controller.go#L210-L222), so even if the Component clean up "fails" due to a missing GitOps repository, the deletion of either resources won't be blocked
> 
> FYI the PR is basically the extra step discussed here: [#451 (comment)](https://github.com/redhat-appstudio/application-service/pull/451#discussion_r1529559854)
> 
> The main reason is that we want to fix the component deletion metrics. I've tried to revisit the investigation on the component deletion errors after your comment. My notes:
> 
> * The truth is In the latest error logs I cannot see any `Unable to update GitOps repository for component..` which comes from an unsuccessful deletion.
> * As we have already noticed the main log is: `Operation cannot be fulfilled on...the object has been modified;`. I think this error comes from [line#213](https://github.com/redhat-appstudio/application-service/blob/main/controllers/component_controller.go#L213) where we are trying to update the `to-be-deleted` component.
> * I can see that in [line#194](https://github.com/redhat-appstudio/application-service/blob/main/controllers/component_controller.go#L194) we are increasing the total deletion attempts. But we only increase the component deletion failed/succeeded for specific cases (see [line#205](https://github.com/redhat-appstudio/application-service/blob/main/controllers/component_controller.go#L205) & [line#219](https://github.com/redhat-appstudio/application-service/blob/main/controllers/component_controller.go#L219)).
> * Having in mind the above note I think the reason of the component deletion not successful alert might be coming because our total attempts are much bigger than the sum of failed + succeeded (e.g we increase the total attempts but we don't equally increase the failed/successful). As a result the succeeded attempts should be only a small pertecantge of the total deletion attempts.
> 
> From the above notes I'd say we could:
> 
> * Increase total attempts only for cases we decide that the attempt is failed/succeded
>   OR
> * We add succeded attempts for lines [214](https://github.com/redhat-appstudio/application-service/blob/main/controllers/component_controller.go#L214) & [216](https://github.com/redhat-appstudio/application-service/blob/main/controllers/component_controller.go#L216)
> 
> wdyt? cc @maysunfaisal @johnmcollier

as discussed I'm closing for now the PR. If in the future this solution is needed we can re-open.

PR #461, opened 2024-04-03
Update OWNERS file

maysunfaisal:
Update OWNERS file

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -3,7 +3,6 @@
 approvers:
 - elsony
 - johnmcollier
-- kim-tsao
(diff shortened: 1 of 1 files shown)


maysunfaisal:
@thepetk could you help review?

thepetk (approved):
/lgtm I'm not sure only if we need to add Jdubrick too here :thinking:

maysunfaisal:
@thepetk I think for now we can keep it as ppl who have contributed to the repo

thepetk:
> for

Sgtm

PR #475, opened 2024-05-15
[KFLUXBUGS-1290] Handle ownerrefs in webhook

johnmcollier:
### What does this PR do?:
Updates the Component webhooks to add in logic to set the owner references. The defaulting webhook will check to see if the Component has the owner reference set every time the resource is created/updated, and if not set, it will update the resource to have it set.

https://issues.redhat.com/browse/KFLUXBUGS-1290

### PR acceptance criteria:

- [ ] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact

### DIFF
--- a/controllers/webhooks/component_webhook.go
+++ b/controllers/webhooks/component_webhook.go
@@ -55,8 +55,7 @@ func (w *ComponentWebhook) Register(mgr ctrl.Manager, log *logr.Logger) error {
 
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 
-// +kubebuilder:webhook:path=/mutate-appstudio-redhat-com-v1alpha1-component,mutating=true,failurePolicy=fail,sideEffects=None,groups=appstudio.redhat.com,resources=components,verbs=create;update,versions=v1alpha1,name=mcomponent.kb.io,admissionReviewVersions=v1
-// +kubebuilder:webhook:path=/mutate-appstudio-redhat-com-v1alpha1-component,mutating=true,failurePolicy=fail,sideEffects=None,groups=appstudio.redhat.com,resources=components/status,verbs=create;update,versions=v1alpha1,name=mcomponent.kb.io,admissionReviewVersions=v1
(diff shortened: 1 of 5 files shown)


controllers/webhooks/component_webhook.go
```diff
@@ -54,10 +55,34 @@ func (w *ComponentWebhook) Register(mgr ctrl.Manager, log *logr.Logger) error {
...
+	component := obj.(*appstudiov1alpha1.Component)
+	compName := component.Name
+	componentlog := r.log.WithValues("controllerKind", "Component").WithValues("name", compName).WithValues("namespace", component.Namespace)
+
+	if len(component.OwnerReferences) == 0 && component.DeletionTimestamp.IsZero() {
+		// Get the Application CR
+		hasApplication := appstudiov1alpha1.Application{}
+		err := r.client.Get(ctx, types.NamespacedName{Name: component.Spec.Application, Namespace: component.Namespace}, &hasApplication)
+		if err != nil {
+			// Don't block if the Application doesn't exist yet - this will retrigger whenever the resource is modified
```
maysunfaisal, line 71: what happens when in the first pass of this code, Application does not exist and Component create proceeds but... Next course of action is Application Delete?

johnmcollier: Hmm, good question.

I think you're right that there is a small gap where that could happen, but I'm not entirely sure how to avoid it, if we want to handle the owner references in the webhooks? I don't think returning an err here is feasible since we don't want to block resource creation/updates.

I think even in the old approach when we handled it in the controller, this could happen if Application deletion occurred before [this block](https://github.com/redhat-appstudio/application-service/blob/f00d9c392bf901acbb78b1e9c672d8958589331b/controllers/component_controller.go#L174-L189).

WDYT?

maysunfaisal: I think the alternative is when an Application is deleted,
1. if there is a possibility to check for owner references to it and if there are none at all, then add them in to get deleted. Sounds tedious. 
2. it reconciles Components, if there is no update, Component reconciler will exit quickly but it may add in the owner references, but i think this may potentially introduce a race condition..

At this stage, i think its too much work.. I will leave it up to you..

/lgtm

maysunfaisal (approved):

PR #481, opened 2024-06-03
Remove HAS controllers

johnmcollier:
- Removes HAS controllers as outlined in https://issues.redhat.com/projects/DEVHAS/issues/DEVHAS-647
- Removes packages that are now unused
- Removes CDQ go module
- Removes pact tests
- Removes obsolete github actions
- Removes the tini binary build from the Dockerfile
- Moves webhooks into their own package under the root of the repository
- Updates main.go to remove the controller invocations, and update webhook package


maysunfaisal (approved):
sad to see it go but
/lgtm

